PR #3, opened 2021-09-08
Ensure LaunchAgents directory exists

sbkg0002:
On fresh OSX installs, the `"$HOME/Library/LaunchAgents/"` folder does not exist.


abiosoft:
Thanks for this.

PR #8, opened 2021-09-11
feat: added compose cli on Mac Host

mmornati:
## Why?
The latest version of `docker-compose` is proposed as docker CLI plugin, available as `docker compose`.
Docker-Desktop is already providing this plugin by default and having it configured here will provide the same developer experience.

## What?
Added a new `start` parameter which is configuring on the Mac Host machine the Compose CLI by default.
To install during the machine provisioning (or even to add it to an existing machine), just run:
```
colima start --with-compose-cli
```
Once installed you can simply access to docker-compose using:
```
docker compose <COMMAND>
```

The `docker-compose` script is still available and it can be installed on your machine too (`brew install docker-compose`).

### DIFF
--- a/colima
+++ b/colima
@@ -21,6 +21,7 @@ export MEMORY="4"
 export DISK="60"
 export CONF_CHANGED=""
 export ENABLE_KUBERNETES=""
+export ENABLE_COMPOSE_CLI=""
 
 # init
 mkdir -p "$CONF_DIR"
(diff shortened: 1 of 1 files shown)


abiosoft:
Hi @mmornati, thanks for your interest in the project and your efforts on putting this together. It is very well appreciated.

However, I would have to decline this. I do not agree with this implementation but more importantly, this contribution is not in accordance with the immediate plans of the project.

The project is attracting some interest and as stated in the docs, Colima will soon be rewritten in Go. The bash version was meant to be a working proof of concept.  For now, only critical fixes (or features) to the current bash version would be added.

I do understand that having docker-compose provided out of the box would be more in line with what people are used to with Docker Desktop. It is being considered.

Thanks.

PR #17, opened 2021-10-08
Add --mount and --mount-rw option to configure additional host directories to expose  to guests (as writable).

k1LoW:
Hi @abiosoft. Thanks for providing a great tool.

I added the `--mount` option because when I use Colima, I often manually modify the `mounts:` in the generated lima.yml.

### Usage

```
$ colima start --cpu 8 --memory 8 --disk 128 --mount '~/src' --mount '~/tmp'
```

---

I know that [Colima is planned to be rewritten in Go](https://github.com/abiosoft/colima#project-goal).

Even if you can't merge this pull request, I'd be happy to get your feedback on the `--mount` option 😄

### DIFF
--- a/colima
+++ b/colima
@@ -21,7 +21,7 @@ export DISK="60"
 export CONF_CHANGED=""
 export ENABLE_KUBERNETES=""
 declare -a MOUNTS=()
-declare -a RW_MOUNTS=()
+declare -a MOUNTS_RW=()
 
 # derived values
 vm_user() (lima whoami)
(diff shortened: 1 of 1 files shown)


k1LoW:
@abiosoft @ivorodrigues Thank you for the nice feedback.

I changed the location added by the mount option to read-only and added a new `--rw-mount` option.

abiosoft:
@k1LoW that's the penalty for using pre-release software :D.

Kindly edit ~/.colima/colima.yaml and add `sshPort: 41122` under `vm` config and it should work fine afterwards. Or if the vm state is not critical, you can delete and recreate afresh.

Nonetheless, I'll add a fix to fall back to the default port for rare cases like yours.

Thanks.

PR #27, opened 2021-10-18
feature: podman runtime

hown3d:
Adds support for an additional runtime: podman.


hown3d:
Hey @abiosoft, updated this PR for the new alpine support. 
PTAL.

I splitted the alpine versions into 3 editions, but im not quite happy with the hardcoded path and digest 😕

abiosoft:
Thanks, will review this soon.
It will be nice to have this as part of v0.3.0.

environment/container/podman/podman.go
```diff
@@ -0,0 +1,126 @@
...
+				return fmt.Errorf("Error running rootfull podman: %w", err)
+			}
+			p.guest.Run("sleep", "5")
+			// set permissions of rootfull podman socket to user for accessing via ssh
+			// docker has the docker group which podman doesnt have :/
+			user, err := user.Current()
+			if err != nil {
+				return fmt.Errorf("Couldn't read current user: %w", err)
+			}
+			return p.guest.Run("sudo", "chown", user.Name+":"+user.Name, "-R", "/run/podman/")
```
pjoomen, line 77: On my system this results in:
```
> chown: unknown user/group Pepijn Oomen:Pepijn Oomen
```

Changing to `user.Username` (ref. https://pkg.go.dev/os/user#User) resolves this.

pjoomen (changes requested):

environment/container/podman/podman.go
hown3d: Thanks for the fix, but maybe the best solution would be using uid and gid, since they're unlikely to change.

jandubois: Note that lima will use `lima` as the user name in the guest if the host user name is not a valid Linux user name, so quoting it will not actually help: https://github.com/lima-vm/lima/blob/master/pkg/osutil/user.go#L83-L101

Lima will make the UID in the guest match the UID on the host, but the GID is not predictable and must be looked up at runtime.

hown3d: If the Uid matches, I think it's not neccesary to set the Gid as Group, since 99% of the time, the VM will be used with only 1 user.

So IMO, setting only the Uid should be fine

abiosoft:
Definitely not this PR. But there are plans to add podman.

abiosoft:
@hown3d hey, sorry if you misunderstood my message https://github.com/abiosoft/colima/pull/27#issuecomment-1145552247. 

What I meant was that there are lots of changes already and due to that, this PR cannot be merged. The changes needed would be better off in another PR.

Your efforts are much appreciated.

jedahan:
> Definitely not this PR. But there are plans to add podman.

is this the best issue to watching to follow podman support, or is there a better place?

aparcar:
@abiosoft did the work continue anywhere else?

abiosoft:
@aparcar @jedahan not at the moment, are you primarily interested in the podman runtime?

aparcar:
Yes, I'm already using it within the coloma VM since it's running faster than the official podman VM. However it requires some manual work which is Leipzig like to avoid

Issue #30, opened 2021-10-19
Boot log only shows the last lines


cmoulliard:
## Suggestion

When we create a new VM using `colima start`, then the client displays the log about the commands executed but only a few lines are displayed and rotation take place after x lines.

Would it be possible to have an option to see all the commands executed when the VM is created ?

```
colima start --runtime containerd --mount $HOME/temp:w
INFO[0000] starting colima
INFO[0000] creating and starting ...                     context=vm
> msg="[hostagent] time=\"2021-10-19T13:11:45+02:00\" level=info msg=\"Using cache \\\"/Users/cmoullia/Library/Caches/lima/download/by-url-sha256/a1ebbffd3dcb629755d3f6bINFO[0053] starting ...                                  context=containerd
INFO[0054] done
```

PR #31, opened 2021-10-19
feat: verbose output

hown3d:
This PR tries to solve #30.
It adds a flags --verbose which sets the output of commands to stdout instead of terminal.VerboseWriter.

Usage: `colima --verbose <command>`


abiosoft:
@hown3d Thanks for this! Much appreciated.

I would say we should try to get this merged before https://github.com/abiosoft/colima/pull/27.

PR #56, opened 2021-11-09
feature: Support SSH agent forwarding

shioshiota:
I implemented a feature to support SSH agent forwarding like Docker Desktop for Mac.
https://docs.docker.com/desktop/mac/networking/#ssh-agent-forwarding

I tested this feature on a machine with Apple M1 Max processor.
```
shiota@lima-colima:~$ ssh-add -l
Could not open a connection to your authentication agent.
shiota@lima-colima:~$ SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock ssh-add -l
256 SHA256:yz69smEOak6d+q+X8I3pMtNVa/yu6kegZlVT3zRZYt8 shiota@TetsuyanoMBP.AirPort (ED25519)
```

PR #64, opened 2021-11-17
When SSH_AUTH_SOCK is set to a socket on the host, forward that to the VM.

scr-oath:
This is most useful for git access, but certainly communicating with other ssh-able destinations is handy.

When creating a docker, one can pass `-e SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock` to access their ssh-agent.

### DIFF
--- a/environment/container/docker/socket.go
+++ b/environment/container/docker/socket.go
@@ -3,10 +3,11 @@ package docker
 import (
 	_ "embed"
 	"fmt"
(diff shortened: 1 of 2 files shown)


abiosoft:
Thanks for putting this together. Looks like this is an essential feature for many as someone made a similar PR recently https://github.com/abiosoft/colima/pull/56.

As I did with the other PR, I will be closing this as well for same reasons.

Thanks.

Issue #83, opened 2021-12-07
Cannot chown or chmod bind mounted files/dirs


chrisdostert:
Hello and thank you for the awesome project!

I'm running into an inconsistency between docker for mac and lima. 
With docker for mac, I can successfully chown & chmod bind mounted files/dirs within a container.

If I attempt this with colima I get `chown: file.txt: Permission denied`

thejan2009:
Related upstream issue: https://github.com/lima-vm/lima/issues/231

lorsatti:
I can `chmod` (tested with colima 0.3.0, but don't know if it is related). `chown` still not working.

jklewa:
Getting this issue on some popular images, which will attempt to correct host-mounted volume file ownership/permissions during their entrypoint scripts
```
postgres:/var/lib/postgresql/data
redis:/data
```

HansG89:
Same problem with mosquitto mqtt broker from [dockerhub](https://hub.docker.com/_/eclipse-mosquitto/)

`mosquitto:/mosquitto`

jhogendorn:
Is this considered fixed or still outstanding? I'm experiencing this issue with the jekyll/jekyll container, on the HEAD install of colima, lima, qemu.

I've tried following these instructions: https://github.com/lima-vm/lima/issues/231#issuecomment-1145113879

But have not had any luck.

abiosoft:
@jhogendorn what is your exact situation? Have you tried starting Colima with `--mount-type 9p`?

jhogendorn:
> @jhogendorn what is your exact situation? Have you tried starting Colima with `--mount-type 9p`?

I'm running a jekyll container with docker compose.

```
version: '2.2'

services:
  jekyll:
    image: jekyll/jekyll:stable
    command: jekyll serve --watch --force_polling --verbose
    ports:
      - 4000:4000
    volumes:
      - .:/srv/jekyll
```

Before, i was getting just a big list of chown errors.

If i start colima with `--mount-type 9p` then i get a slightly different permission error: `/usr/local/lib/ruby/3.1.0/fileutils.rb:243:in `mkdir': Permission denied @ dir_s_mkdir - /srv/jekyll/.jekyll-cache (Errno::EACCES)`

I got around that by running `mkdir .jekyll-cache`

Then it was failing due to https://github.com/envygeeks/jekyll-docker/issues/335

[...]

Issue #54, opened 2021-11-07
Issue with writable volumes


matthieugouel:
Hi 👋

I just came across this project and it seems very promising, thanks for your work! 

I have a little issue while setup a docker container with writable volume (it's a database).

I start Colima and I mount the local directory: 

```
colima start --mount /Users/matthieugouel/Documents/Code:rw
```

Then, I try to create the docker container: 

```
 docker run  --name clickhouse -v $(pwd)/volumes/clickhouse:/var/lib/clickhouse -p 9000:9000 yandex/clickhouse-server:latest
```

But the issue is: 

```
chown: changing ownership of '/var/lib/clickhouse/': Read-only file system
```

It's strange because when I ssh inside the VM I have these rights for the volume directory: 

```
drwxr-xr-x 1 matthieugouel dialout  352 Nov  7 13:45 volumes
```

Would you have an idea of why this issue is happening? 

Best, 
Matthieu.

abiosoft:
It is `:w` not `:rw`.

matthieugouel:
In this case I have:

```
chown: changing ownership of '/var/lib/clickhouse/': Permission denied
```

dustinrue:
This might be because the directory exists inside the colima VM. Probably the safest thing is to delete the colima VM you created and create it again with the corrected `:w` option

matthieugouel:
I cleaned up everything with:

```
colima stop
colima delete
rm -Rf volumes/clickhouse
```

Then I started again the VM with: 

```
colima start --mount /Users/matthieugouel/Documents/Code:w
```

I ssh onto the VM to check the state of the directories. At this point the directory volumes is there with the right permissions and the directory clickhouse is not there as intended. 

I then created the clickhouse container:

```
docker run  --name clickhouse -v $(pwd)/volumes/clickhouse:/var/lib/clickhouse:rw -p 9000:9000 yandex/clickhouse-server:latest
```

I get the error: 

```
docker: Error response from daemon: error while creating mount source path '/Users/matthieugouel/Documents/Code/research/iris/volumes/clickhouse': chown /Users/matthieugouel/Documents/Code/research/iris/volumes/clickhouse: permission denied.
```

If I check on the VM the directory has been created:

```
drwxr-xr-x 1 matthieugouel dialout   64 Nov  8 11:19 clickhouse
```

If I start again without touching anything I get simply this error:
```
chown: changing ownership of '/var/lib/clickhouse/': Permission denied
```

[...]

PR #89, opened 2021-12-12
add documenation for installation via nix

aaschmid:
Introduced in `nixpkgs` by https://github.com/NixOS/nixpkgs/pull/150282
findable with https://search.nixos.org/packages?channel=unstable&show=colima&from=0&size=50&sort=relevance&type=packages&query=colima.

**Note:** It will take another 5-7 days that package will appeare in `nixpkgs-unstable`. However from than it will be automatically included into the next major nix versions.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -14,6 +14,20 @@ Development Version
 brew install --HEAD colima
 ```
 
+## Nix
+
+Only stable Version
+
+```
+nix-env -i colima
+```
+
+Or using solely in a `nix-shell`
+
+```
+nix-shell -p colima
+```
+
 ## Binary
 
 Binaries are available with every release on the [releases page](https://github.com/abiosoft/colima/releases).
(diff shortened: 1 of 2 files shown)


abiosoft:
@aaschmid thanks a lot for this.

Is there a way to support installing bleeding edge version like homebrew? Considering that project is still early and in active development, there are chances that the current unreleased version is far ahead of the last stable release, and some users would not mind that.

abiosoft (approved):

aaschmid:
@abiosoft: Yes, there is a way but this is very `nix` specific. I can document this, however, every nix should now how to override versions...

aaschmid:
I subscribed to new releases and will asap update the `nixpkgs` with ;-)

(Side note: The biggest issue currently might be #54, #64, #83 which depend on https://github.com/lima-vm/lima/issues/231 afik)

PR #104, opened 2022-01-11
Add GolangCI Lint

kensipe:
* Adding GolangCI Lint for linting
* Added lint task to Makefile
* Resolved basic linting issues

## Lint Resolution

* Removed `kubernetesVersion` as it is private and used.  If desired, please indicate how it is to be used 
* `time.Since` is preferred to `time.Now().Sub` otherwise same functionality
* Likely the most controversial, The assignment of `downloadURL := u` was never used.  if err not nil, it is assigned to `u` otherwise an error is returned.  It is better and more readable in Go to not use `else` unless required, in this case, it is much more clear.

Looking forward to adding more linting rules.   I was looking for feedback on this first.
 
Signed-off-by: Ken Sipe <kensipe@gmail.com

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,6 +17,9 @@ jobs:
         with:
           go-version: 1.17
 
+      - name: Lint
+        run: make lint
+
       - name: Build
         run: go build -v ./...
 
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -17,8 +17,11 @@ jobs:
         with:
           go-version: 1.17
 
-      - name: Lint
-        run: make lint
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+            version: v1.43.0
+            args: --timeout 3m0s
 
       - name: Build
         run: go build -v ./...
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,6 @@ GOARCH_x86_64 = amd64
 GOARCH_aarch64 = arm64
 GOARCH_arm64 = arm64
 GOARCH ?= $(shell echo "$(GOARCH_$(ARCH))")
-GOLANGCI_LINT_VER = "1.43.0"
 
 all: build 
 
@@ -28,9 +27,5 @@ install:
 	chmod +x /usr/local/bin/colima
 
 .PHONY: lint
-lint: ## Run golangci-lint
-ifneq (${GOLANGCI_LINT_VER}, "$(shell ./bin/golangci-lint version --format short 2>&1)")
-	@echo "golangci-lint missing or not version '${GOLANGCI_LINT_VER}', downloading..."
-	curl -sSfL "https://raw.githubusercontent.com/golangci/golangci-lint/v${GOLANGCI_LINT_VER}/install.sh" | sh -s -- -b ./bin "v${GOLANGCI_LINT_VER}"
-endif
-	./bin/golangci-lint --timeout 3m run
+lint: ## Assumes that golangci-lint is installed and in the path.  To install: https://golangci-lint.run/usage/install/
+	golangci-lint --timeout 3m run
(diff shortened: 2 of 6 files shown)


kensipe:
@abiosoft any thoughts? or help regarding Linting?

abiosoft:
@kensipe I was working on the codebase at the time and would rather not deal with conflicts.
You can rebase now and I'll have a look afterwards.

Thanks.

kensipe:
it's worth noting but perhaps obvious... the first lint requires connectivity to pull the versioned linter.  All other lints can work with a disconnected machine.

Makefile
```diff
@@ -25,3 +26,11 @@ build:
 install:
 	cp _output/binaries/colima-$(OS)-$(ARCH) /usr/local/bin/colima
 	chmod +x /usr/local/bin/colima
+
+.PHONY: lint
+lint: ## Run golangci-lint
```
abiosoft, line 31: I would prefer to assume the environment has golangci-lint installed, rather than have the installer in the Makefile.

kensipe: I can make that change... however the issue with that is each contributor will likely have a different version of golangci-lint or will over time... In my experience it has lead to PRs to fix linting which were newly added.  It also seems valuable to be able to lint in CI and have a standard for identifying which version is the standard.  Do you have a recommendation on how to manage that?  or willing to deal with the consequences?   and how do you expect the CI to define the linter version?

environment/container/kubernetes/kubernetes.go
```diff
@@ -50,9 +50,6 @@ func (c kubernetesRuntime) Running() bool {
 func (c kubernetesRuntime) runtime() string {
 	return c.guest.Get(environment.ContainerRuntimeKey)
 }
-func (c kubernetesRuntime) kubernetesVersion() string {
```
abiosoft, line 53: This is actually temporary and it will soon be used again. Can we temporarily disable linting for this?

Makefile
kensipe: ah... actually I think there was an addition to the linter config to specify the version to be expected... please hold :)

kensipe: didn't find it... the consequence remains.   I will remove based on your comment but IMO it will result in checks working on different machines differently and leaves the concern of how to manage the CI linter version

kensipe: it also forces all projects for a developer to be at the same version of linting which is not desirable or expected

kensipe:
@abiosoft ok.  Add adjustments made.  Worth noting that this is the initial stage.  The config file is empty making this PR small.  I was planning on adding a number of linters for your review after this is merged.

As noted, I still have concerns over linter versioning but happy to seeing linting 

As note that based on guidance, there is a total assumption the linter is in the path.  The result of it not being in the path is:

```
make lint
golangci-lint --timeout 3m run
make: golangci-lint: No such file or directory
make: *** [lint] Error 1
```

abiosoft:
One last thing to make this actually be part of the CI, can you add a `lint` step as part of the build job in `.github/workflows/go.yml`?

abiosoft:
> As note that based on guidance, there is a total assumption the linter is in the path. The result of it not being in the path is:
> ```
> make lint
> golangci-lint --timeout 3m run
> make: golangci-lint: No such file or directory
> make: *** [lint] Error 1
> ```

Not an issue, it will be available in the CI environment. And anyone that wants to run it locally can always install it.

kensipe:
@abiosoft ✅

.github/workflows/go.yml
```diff
@@ -17,6 +17,9 @@ jobs:
         with:
           go-version: 1.17
 
+      - name: Lint
```
abiosoft, line 20: We will need a prior step to install the linter, similar to how Go was installed.

abiosoft: A step like this can be added.

``` yaml
      - name: golangci-lint
        uses: golangci/golangci-lint-action@v2
        with:
          version: v1.29
```

From https://github.com/golangci/golangci-lint-action#how-to-use

kensipe: ha... right!    I have deep exp with circleci, travis, jenkins etc.   but new to github workflows...   it's worth a eagle eye review.   The docs recommend a separate workflow job which I agree with. (jobs are run in parallel).  But configured based on guidance.   Ready!

PR #130, opened 2022-01-18
fix(readme): add docker-desktop migration

georgettica:
the FAQ says it can run alongside, but not how to switch between them

this shows the way

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,14 @@ Docker for Mac. You should run either, not both.~~
 
 Yes, from version v0.3.0, Colima can run alongside Docker for Mac.
 
+to switch between them, use `docker contexts`
+```shell
+$ docker context list
+NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                              KUBERNETES ENDPOINT   ORCHESTRATOR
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,6 +142,14 @@ Docker for Mac. You should run either, not both.~~
 
 Yes, from version v0.3.0, Colima can run alongside Docker for Mac.
```
abiosoft, line 143: I would prefer to simply change the existing text to something like


> Yes, from version v0.3.0 Colima leverages Docker contexts and can thereby run alongside Docker for Mac.
>
> `docker context list` can list all contexts and `docker context use` can be used to change the active context.

README.md
```diff
@@ -142,6 +142,14 @@ Docker for Mac. You should run either, not both.~~
...
 
+to switch between them, use `docker context`
```
abiosoft, line 145: I would prefer to simply change the existing text to something like


> Yes, from version v0.3.0 Colima leverages Docker contexts and can thereby run alongside Docker for Mac.
>
> `docker context list` can list all contexts and `docker context use` can be used to change the active context.

abiosoft (approved):

Issue #24, opened 2021-10-14
Use `goreleaser` to create releases


luisdavim:
I'd like to suggest using https://goreleaser.com/ to cut releases, this will allow you to easily create  cross-compiled versions for Intel and ARM macs.

mavogel:
Or just make a PR, declare it as WIP and link the issue here :) Then folks know and won't do double work

mavogel:
feel free to close then as well @abiosoft

developer-guy:
Ah sorry about that, I missed that one 🤦

developer-guy:
@abiosoft can you please re-open this issue, I'm going to complete it tonight

mavogel:
unfortunately it's not necessary any more @developer-guy . See #147

PR #147, opened 2022-01-27
ci: add goreleaser

mavogel:
Hi, happy to contribute for #24  :) 

do I get it right that your `aarch64` build is arm for linux?

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,7 +24,10 @@ jobs:
             args: --timeout 3m0s
 
       - name: Build
-        run: go build -v ./...
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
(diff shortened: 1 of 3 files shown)


abiosoft:
I actually do not see the need for goreleaser as I have the releases pretty much sorted. 
If you do not mind, what benefit does this PR offer?

> do I get it right that your aarch64 build is arm for linux?

Yeah, that's accurate.

abiosoft:
> Hi, it would offer a more standardized and extensible way to release go binaries and libraries. Based on your comment: [#24 (comment)](https://github.com/abiosoft/colima/issues/24#issuecomment-953822544) you we're open for it.

😄 yeah, I was open to it at the time. However, due to lack of progress from the original volunteer, I have disregarded it handled releases with typical github workflows.

PR #174, opened 2022-02-09
Prepare for v0.3.3

### LINKED ISSUE #171: Unable to login on AWS registry when using `containerd`
First of all, thank you for bringing alternative to Docker Desktop 😌 It works like a charm =)

But, I have a problem :p 

When using colima with `containerd` runtime, I cannot login to the AWS registry using the `nerdctl login` (that I have aliased to `docker` command). Regular login to the "Official" Docker registry works fine.

The issue seems discuss and resolved on [nerdctl](https://github.com/containerd/nerdctl/issues/429). So is it about a dependencies update or I miss something?

I've tried running `lima` directly with nerdctl and it works (I guess the ubuntu image for containerd is more stuffed than the alpine used by colima).

```
> colima start --runtime=containerd
> colima nerdctl install --path "${HOME}/opt/bin/docker"
> aws --profile "shared-services-aws" ecr get-login-password --region "${AWS_REGION}" | docker login --username "AWS" --password-stdin "${AWS_SHARED_SERVICES_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com"

time="2022-02-09T08:30:34Z" level=error msg="failed to call tryLoginWithRegHost" error="unexpected status code 404" i=0
time="2022-02-09T08:30:34Z" level=fatal msg="unexpected status code 404"
```

abiosoft:
Fixes #171: update nerdctl to 0.16.1
Core: add configurable dns with `--dns` flag.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -26,7 +26,8 @@ The --runtime, --disk and --arch flags are only used on initial start and ignore
 		"  colima start --with-kubernetes\n" +
 		"  colima start --runtime containerd --with-kubernetes\n" +
 		"  colima start --cpu 4 --memory 8 --disk 100\n" +
-		"  colima start --arch aarch64",
+		"  colima start --arch aarch64\n" +
+		"  colima start --dns 1.1.1.1 --dns 8.8.8.8",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		return newApp().Start(startCmdArgs.Config)
 	},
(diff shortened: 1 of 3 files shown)


environment/container/docker/docker.go
```diff
@@ -119,3 +126,34 @@ func (d dockerRuntime) Version() string {
...
+			return nil
+		}
+
+		// we are utilising the host cache path as it is the only guaranteed mounted path.
+
+		dockerCertsCacheDir := filepath.Join(config.CacheDir(), "docker-certs")
+		if err := d.host.RunQuiet("mkdir", "-p", dockerCertsCacheDir); err != nil {
+			return err
+		}
+
```
eljohnson92, line 146: I tried testing this locally, and I see the directory gets created, but none of the subdirectories/files get populated. I might be misreading the code here, but it looks like you create the directory in the cache on the host, but never copy the host subfolders into that directory?

abiosoft: You're actually right, thanks for spotting.

abiosoft: fixed https://github.com/abiosoft/colima/releases/tag/v0.3.3-pre-02

PR #182, opened 2022-02-15
docs: add documentation for MacPorts installation

runar:
Colima is available on [MacPorts](https://ports.macports.org/port/colima/), so the relevant docs should be updated with installation instructions.


abiosoft (approved):

abiosoft:
Thanks :)

Issue #200, opened 2022-02-24
"error provisioning docker" when override.yaml exists


Jdban:
I am running `colima` on an M1 mac. 
I have tried with the latest released version of `colima`, as well as the `HEAD` version, both have the same issue when using an `override.yaml` for lima.

I need to change some settings in [override.yaml](https://github.com/lima-vm/lima/blob/01ce318b113b88196ad4862278fdb9876cb03cc8/pkg/limayaml/default.yaml), so I have it saved to `~/.lima/_config/override.yaml`. 
I have yet to make any changes to the `override.yaml` file, it is still identical to the `default.yaml` file.

I've tried a couple different revisions of `default.yaml` from the lima repo since it seems to have changed a bit recently (just in case it changed behavior) but it did not. 

Whenever the override.yaml has values in it, I get an error about "group docker does not exist".

```bash
➜  colima start --cpu 1
INFO[0000] starting colima                              
INFO[0000] creating and starting ...                     context=vm
INFO[0038] provisioning ...                              context=docker
>  usermod: group 'docker' does not exist
FATA[0038] error provisioning docker: error at 'provisioning': error fixing user permission: exit status 6 

➜ colima status       
INFO[0000] colima is running                            
INFO[0000] runtime: docker                              
INFO[0000] arch: aarch64           
                     
➜  docker ps
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
````

If i delete the `override.yaml` (or comment out the entire file), then everything seems to work just fine

```bash
➜  colima start --cpu 1
INFO[0000] starting colima                              
INFO[0000] creating and starting ...                     context=vm
INFO[0021] provisioning ...                              context=docker
INFO[0022] restarting VM to complete setup ...           context=docker
INFO[0022] stopping ...                                  context=vm
INFO[0027] starting ...                                  context=vm
INFO[0048] starting ...                                  context=docker
INFO[0053] waiting for startup to complete ...           context=docker
INFO[0053] done                
                         
➜  docker ps
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
```

Is this a bug or am I doing something wrong?

abiosoft:
Did you override the image iso as well? If yes, that is the problem.

Jdban:
> Did you override the image iso as well? If yes, that is the problem.

I don't believe I did anything like that. How can I verify this?

abiosoft:
[...]

PR #202, opened 2022-02-25
Update Readme to include more information about lima's override.yaml

Jdban:
I figured some more documentation here might keep someone else from running into the same issue that I did here https://github.com/abiosoft/colima/issues/200

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,9 @@ Since Lima is aka Linux on Mac. By transitivity, Colima can also mean Containers
 
 Colima is basically a higher level usage of Lima and utilises Lima to provide Docker, Containerd and/or Kubernetes.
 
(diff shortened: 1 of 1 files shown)


jandubois:
I understand where this is coming from, but the basic idea of copying `examples/default.yaml` into `_config/override.yaml` is just wrong. `override.yaml` is a mechanism to very selectively override a few settings to adapt to your specific setup.

It should not be filled with a bunch of default settings. For most users that file shouldn't even need to exist.

Similarly `_config/default.yaml` should not include all available settings; you should only include things where you want the global lima defaults to be different from what has been compiled into `limactl`.

Otherwise it will be very hard to maintain these files. And if you run into problems, it will be hard for others to help you, if they have to work their way through all your redundant settings files first.

Jdban:
Got it, that makes sense. I'll close this for now and reopen with different text

Issue #192, opened 2022-02-20
Support native volume mount on Mac with 9pfs


srdcarrion:
Just wondering if anyone has looked into supporting 9pfs per:

https://github.com/containers/podman/issues/8016#issuecomment-1034091359 

Really like Colima and have been working around the issues sshfs presents.

kingbuzzman:
Holly crap! Thats A LOT faster, no comparison!

abiosoft:
There has been progress on this upstream on Lima and qemu, should make it into colima soon.

PR #237, opened 2022-03-23
WIP: Support for mount type (reverse-sshfs, 9p)

matejsp:
Initial support for mount-type. It is not default since it is not faster in my case.
https://github.com/abiosoft/colima/issues/192

### DIFF
--- a/environment/vm/lima/yaml.go
+++ b/environment/vm/lima/yaml.go
@@ -124,6 +124,10 @@ func newConf(ctx context.Context, conf config.Config) (l Config, err error) {
 		)
 	}
 
+	if conf.VM.MountType != "" {
+		l.MountType = conf.VM.MountType
+	}
+
 	if len(conf.VM.Mounts) == 0 {
 		l.Mounts = append(l.Mounts,
 			Mount{Location: "~", Writable: true},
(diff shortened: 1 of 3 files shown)


abiosoft:
Judging by your comment https://github.com/lima-vm/lima/pull/726#issuecomment-1068817738, looks like is slightly slower with `fscache` enabled. But if that comes with better stability (and filesystem permissions) it might be the better option for the default.

matejsp:
True, but fscache is not enabled by default for now. With fscache disabled it is a lot slower (6x times in case of a lot of small files): https://github.com/lima-vm/lima/pull/726#issuecomment-1067923276

Current sshfs is quite stable for me (I didn't have any problems) and performance matter to me.
But test it yourself. They have the same objection for not having 9p by default. 
Since my disappointment in 9p I am rooting for virtiofsd solution :)

This is initial patch that I used for testing their branch. Since it is not all around best I need to add also cache types and other settings.

matejsp:
I was testing on older master with before latest vmnet fixes. Perhaps I should retest and get another view :)

rfay:
sshfs has pretty amazingly awful problems when you start looking closely... the actual mounted files may be getting into the container long after you've started up with a bind. This doesn't matter much to individual users who are typing along, but it has terrible consequences for automated tests.

environment/vm/lima/yaml.go
```diff
@@ -167,6 +171,7 @@ type Config struct {
 	Memory       string            `yaml:"memory,omitempty"`
 	Disk         string            `yaml:"disk,omitempty"`
 	Mounts       []Mount           `yaml:"mounts,omitempty"`
+	MountType    string            `yaml:"mountType,omitempty"`
```
AkihiroSuda, line 174: The YAML type definition is still under discussion (https://github.com/lima-vm/lima/pull/726)

sebastianrothe:
Seems to be released in lima 0.10.0 https://github.com/lima-vm/lima/blob/master/docs/mount.md#9p

abiosoft:
This is no longer needed as it ended up being part of the other fixes.

Thanks a lot for your efforts.

PR #244, opened 2022-03-28
kubernetes: prevent port forward conflicts for apiserver

milas:
Pick a random free port on the host and use that as the value for
the `--https-listen-port` argument for k3s.

This makes it possible to create multiple colima profiles/instances
with Kubernetes; otherwise, they'll all be listening on 6443, and
only the first one will be port forwarded, so attempting to use any
others will result in x509 errors since the config<>cluster mapping
is erroneous.

### DIFF
--- a/environment/container/kubernetes/k3s.go
+++ b/environment/container/kubernetes/k3s.go
@@ -2,6 +2,7 @@ package kubernetes
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/abiosoft/colima/cli"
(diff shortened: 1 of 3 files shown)


abiosoft:
Actually now with PTP network, it is possible to have multiple instances with Kubernetes enabled since it now binds to the instance IP. This should already be the case if you build from source on `main` branch.

The challenge with this is that the port is not guaranteed to be available on subsequent starts.

milas:
Ahh, nice! I saw the PTP network support but it wasn't working for me at first when building/running locally from `main`.

PR #245, opened 2022-03-28
vmnet: resolve to absolute path for symlink/install

milas:
This is unlikely to affect users who install colima from Homebrew
or similar sources but ensures that the `colima-vmnet` symlink
works even if the `colima` binary is not in `PATH`.


abiosoft (approved):

abiosoft:
Thanks for this

PR #308, opened 2022-05-25
Reverse ln command arguments in FAC.md

bertbaron:
Just wondering if this `ln` command shouldn't be reversed.


abiosoft (approved):
Thanks for spotting it. :)

PR #354, opened 2022-06-27
chore: refactor Makefile

tricktron:
## Changes
- Extract the build logic from build.sh to the Makefile: This makes it easier to override variables to customise the build, e.g we can overwrite `VERSION` or `OUTPUT_DIR` variables more easily. This is especially useful for package manager like nix.
- Uses openssl to calculate the sha256 because it is both available on linux and macOS, so we don't need to check the OS.
- Adds a test rule: Allows to easily run the tests with: `make test`.


abiosoft:
Thanks :)

Makefile
```diff
@@ -8,27 +8,52 @@ GOARCH_aarch64 = arm64
...
 
+.PHONY: vmnet
 vmnet:
 	sh scripts/build_vmnet.sh
 
+.PHONY: install
 install:
-	cp _output/binaries/colima-$(OS)-$(ARCH) /usr/local/bin/colima
-	chmod +x /usr/local/bin/colima
+	mkdir -p $(INSTALL_DIR)
```
abiosoft, line 54: Context: https://github.com/abiosoft/colima/pull/352/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R30
```suggestion
	mkdir -p $(INSTALL_DIR)
	rm -f $(INSTALL_DIR)/$(BIN_NAME)
```

tricktron:
@abiosoft I still need/want to test this with nix. So can you hold on with the merge?

abiosoft:
> @abiosoft I still need/want to test this with nix. So can you hold on with the merge?

That's fine by me. You can also mark the PR as draft and mark it as ready when you're done.

tricktron:
@abiosoft I tested it with nix and it works. Could you run the workflows/tests to check if it does not break anything?

Makefile
```diff
@@ -8,27 +8,53 @@ GOARCH_aarch64 = arm64
...
+.PHONY: fmt
 fmt:
 	go fmt ./...
 	goimports -w .
 
+.PHONY: build
 build:
-	GOOS=$(GOOS) GOARCH=$(GOARCH) OS=$(OS) ARCH=$(ARCH) sh scripts/build.sh
+	go build -ldflags="$(LDFLAGS)" -o $(OUTPUT_DIR)/$(OUTPUT_BIN) ./cmd/colima
+	openssl sha256 -r -out $(OUTPUT_DIR)/$(OUTPUT_BIN).sha256sum $(OUTPUT_DIR)/$(OUTPUT_BIN)
```
abiosoft, line 42: I am hoping the hash file can stay consistent as well.

It is now changed from the following format
```
3fc57d9ca92c802a4a4b0745ffe782e3a1074898530da48802f0097ab36bff58  colima-Darwin-x86_64
```

to the following format
```
9619fa6137350d4060127cd4e241f74f87bda0663bb1077e08e9f62c50ee1579 *_output/binaries/colima-Darwin-x86_64
```

I tried the following but could not get rid of the preceding asterisk `*`.

```suggestion
	cd $(OUTPUT_DIR) && openssl sha256 -r -out $(OUTPUT_BIN).sha256sum $(OUTPUT_BIN)
```

tricktron: @abiosoft
The preceding asterisk has a meaning: it means that the input file is [binary](https://apple.stackexchange.com/questions/230917/need-a-cli-to-check-the-sha256-hash-of-a-file).

Openssl automatically recognises that the colima bin is in binary format whereas shasum does not because it read in text mode by default. So if you want to get rid of the asterisk we need to get back to using `shasum`/`sha256sum` depending on the OS.

abiosoft: If it's a known standard then I am fine with it. 

Thanks.

Makefile
```diff
@@ -8,27 +8,53 @@ GOARCH_aarch64 = arm64
...
 
+.PHONY: fmt
 fmt:
 	go fmt ./...
 	goimports -w .
 
+.PHONY: build
 build:
-	GOOS=$(GOOS) GOARCH=$(GOARCH) OS=$(OS) ARCH=$(ARCH) sh scripts/build.sh
+	go build -ldflags="$(LDFLAGS)" -o $(OUTPUT_DIR)/$(OUTPUT_BIN) ./cmd/colima
```
abiosoft, line 41: Seems the variable does not propagate.

```suggestion
	GOOS=$(GOOS) GOARCH=$(GOARCH) go build -ldflags="$(LDFLAGS)" -o $(OUTPUT_DIR)/$(OUTPUT_BIN) ./cmd/colima
```

PR #438, opened 2022-10-10
feat: add disable-servicelb flag

pmalek:
This PR adds a `kubernetes-servicelb` flag which controls whether to enable service LB in k3s.

This is just a proposal. Ideally we'd introduce a mechanism which would allow to control all the "features" of k3s (as indicated below) via e.g. a slice of strings (flags at the CLI level) which would then be added as `--disable-FEATURE` to k3s server invocation

```
   --disable value                            (components) Do not deploy packaged components and delete any deployed components (valid items: coredns, servicelb, traefik, local-storage, metrics-server)
```

That would though be a breaking change if we wanted to make it consistent and move the ingress controller knob there as well.

Let me know what you think!


abiosoft:
Yeah this works.

I am equally fine with a generic one as well. The only issue iis the inability to validate at startup, but considering this is an advanced feature it should be fine.

The `--ingress` flag is not an issue, it can be deprecated ( or hidden) for the meantime. It should not be removed yet for backward compatibility.

abiosoft:
@pmalek can you kindly confirm if you are still interested in working on a generic flag for k3s features?

pmalek:
@abiosoft Yes, I am interested. In fact you can find my updated proposal in #443. This can be closed in case #443 gets accepted and merged.

abiosoft:
closing in favour of #443

PR #443, opened 2022-10-13
feat: add kubernetes-disable flag

pmalek:
This PR adds a generic `--kubernetes-disable` flag which accepts a slice of names of packaged k3s components to disable.

Full list of those can be obtained by looking at the flags for `k3s server`:

```
   --disable value                            (components) Do not deploy packaged components and delete any deployed components (valid items: coredns, servicelb, traefik, local-storage, metrics-server)
```

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -148,7 +148,7 @@ func init() {
 	startCmd.Flags().BoolVarP(&startCmdArgs.Kubernetes.Enabled, "kubernetes", "k", false, "start with Kubernetes")
 	startCmd.Flags().BoolVar(&startCmdArgs.Flags.LegacyKubernetes, "with-kubernetes", false, "start with Kubernetes")
 	startCmd.Flags().StringVar(&startCmdArgs.Kubernetes.Version, "kubernetes-version", defaultKubernetesVersion, "must match a k3s version https://github.com/k3s-io/k3s/releases")
-	startCmd.Flags().StringSliceVar(&startCmdArgs.Kubernetes.Disable, "kubernetes-disable", nil, "a slice of packaged component names to disable when running with k3s (default empty)")
+	startCmd.Flags().StringSliceVar(&startCmdArgs.Kubernetes.Disable, "kubernetes-disable", nil, "components to disable for k3s e.g. traefik,servicelb (default empty)")
 	startCmd.Flag("with-kubernetes").Hidden = true
 
 	// layer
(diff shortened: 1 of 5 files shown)


cmd/start.go
```diff
@@ -147,7 +148,7 @@ func init() {
 	startCmd.Flags().BoolVarP(&startCmdArgs.Kubernetes.Enabled, "kubernetes", "k", false, "start with Kubernetes")
 	startCmd.Flags().BoolVar(&startCmdArgs.Flags.LegacyKubernetes, "with-kubernetes", false, "start with Kubernetes")
 	startCmd.Flags().StringVar(&startCmdArgs.Kubernetes.Version, "kubernetes-version", defaultKubernetesVersion, "must match a k3s version https://github.com/k3s-io/k3s/releases")
-	startCmd.Flags().BoolVar(&startCmdArgs.Kubernetes.Ingress, "kubernetes-ingress", false, "enable Traefik ingress controller")
+	startCmd.Flags().StringSliceVar(&startCmdArgs.Kubernetes.Disable, "kubernetes-disable", nil, "a slice of packaged component names to disable when running with k3s (default empty)")
```
abiosoft, line 151:
```suggestion
	startCmd.Flags().StringSliceVar(&startCmdArgs.Kubernetes.Disable, "kubernetes-disable", nil, "components to disable for k3s e.g. traefik,servicelb (default empty)")
```

abiosoft:
Thanks for this. Other than minor rewording of the flag doc, I think it's good to go.

pmalek:
Updated the flag description, PTAL.

abiosoft (approved):

ventsislav-georgiev:
Doesn't this change the defaults, now having traefik enabled?

pmalek:
@ventsislav-georgiev It does. We'd have to add traefik to the default slice to retain the backwards compatibility (of some sorts). Is that something we'd like to do?

ventsislav-georgiev:
Well it depends. I am using nginx and having traefik disabled is useful to me. Changing the defaults results in unexpected extra component in my cluster after minor update. However, colima is in alpha version (semver 0.x.x). So breaking changes can be expected at every release. Just wanted to know if it is noted, desired change.

abiosoft:
@ventsislav-georgiev that's a good point. I am in favour in maintaining the backward compatibility when it's easy to do (like this case). Traefik will be disabled by default in a follow-up PR.

Thanks.

Issue #417, opened 2022-09-06
kubernetes version changed when stop/start Colima


prune998:
### Description

Reproducing:

- start colima with a specific k8s version using `colima start --edit` and set version to `v1.25.0-rc1+k3s1` or `colima start --kubernetes-version v1.25.0-rc1+k3s1`
- Validate the version in the config file:
```
grep "  version"  ~/.colima/default/colima.yaml
  version: v1.25.0-rc1+k3s1

k version
...
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.0-rc1+k3s1", ...
```
- stop colima with `colima stop`
- restart colima without any change to the config with `colima start`
- version is changed to default ` v1.23.6+k3s1`
you can see this in the logs:
```
❯ colima start
INFO[0000] starting colima
INFO[0000] runtime: containerd+k3s
INFO[0000] preparing network ...                         context=vm
INFO[0000] starting ...                                  context=vm
INFO[0030] provisioning ...                              context=containerd
INFO[0030] starting ...                                  context=containerd
INFO[0036] provisioning ...                              context=kubernetes
INFO[0036] version changed to v1.23.6+k3s1, downloading and installing ...  context=kubernetes  <-----------
INFO[0048] loading oci images ...                        context=kubernetes
INFO[0056] starting ...                                  context=kubernetes
INFO[0068] done

❯ grep "  version"  ~/.colima/default/colima.yaml
  version: v1.23.6+k3s1
```

### Version

Colima Version:

```
❯ colima version
colima version 0.4.4
git commit: 8bb1101a861a8b6d2ef6e16aca97a835f65c4f8f

runtime: containerd
arch: x86_64
client: v0.20.0
server: v1.5.8

kubernetes
Client Version: v1.25.0
Kustomize Version: v4.5.7
Server Version: v1.23.6+k3s1
```

Lima Version:
```
❯ limactl -v
limactl version 0.11.3
```

Qemu Version: na

### Operating System

- [X] macOS Intel
- [ ] macOS M1
- [ ] Linux

### Reproduction Steps

```
colima start --kubernetes-version v1.25.0-rc1+k3s1
INFO[0000] starting colima
INFO[0000] runtime: containerd+k3s
INFO[0000] preparing network ...                         context=vm
INFO[0000] starting ...                                  context=vm
INFO[0030] provisioning ...                              context=containerd
INFO[0030] starting ...                                  context=containerd
[...]

PR #485, opened 2022-11-18
template: save default config from template

hfqin0:
I found colima can not update the default config when I run `colima template`.
E.g.,
I change the kubernetes version in the template by running `colima tempalte`
image
but the version in the default config does not change when I run `colima start --edit`
image

This PR fixes it.

### DIFF
--- a/cmd/template.go
+++ b/cmd/template.go
@@ -63,6 +63,11 @@ var templateCmd = &cobra.Command{
 
 		log.Println("configurations template saved")
 
+		// save to default config file
+		if err := configmanager.SaveToFile(cf, config.File()); err != nil {
+			return fmt.Errorf("error saving default: %w", err)
+		}
+
 		return nil
 	},
 }


abiosoft:
The template is not meant for modifying the default profile's config. Rather, it is meant to be the initial config for a newly created VM. i.e. if you were to start afresh.

If you need to modify your current config, you can simply `colima start --edit` or edit the config file directly in `~/.colima/default/colima.yaml`.

It is mainly useful for those that spin up multiple instances or frequently teardown the VM.

hftsin:
yes, thanks for your explanation for the `template` command. In my daily use of colima, i want to save the latest config that updated from `--edit`. So do you think it is necessary to update the default config once we edit it?

abiosoft:
oh gotcha. You are likely hitting the bug specific to saving Kubernetes https://github.com/abiosoft/colima/issues/417. 
It was fixed in https://github.com/abiosoft/colima/pull/442 but has not been released yet.

You can `brew install --HEAD colima` to use the development version (that includes the fix) for now, until the next release.

Thanks.

abiosoft:
FYI any changes you make during `--edit` will persist.

hftsin:
> Oh, gotcha. You are likely hitting the bug specific to saving Kubernetes version #417. It was fixed in #442 but has not been released yet.
> 
> You can `brew install --HEAD colima` to use the development version (that includes the fix) for now, until the next release.
> 
> Thanks.

Yes, that is what I need. Thanks for you kind work.

PR #486, opened 2022-11-18
template: save default config from template

hftsin:
I found colima can not update the default config when I run `colima template`.
E.g.,
I change the kubernetes version in the template by running `colima tempalte`
image
but the version in the default config does not change when I run `colima start --edit`
image

This PR fixes it.


abiosoft:
Kindly see my comment on your previously opened PR https://github.com/abiosoft/colima/pull/485#issuecomment-1319800010

PR #510, opened 2022-12-14
fix: detect macos version

hftsin:
## Changes

- rename `MacOS13OrNewer` to make the function more readable
- decrease the number of calling version comparison
- disable duplicated macOS version check in case of M1


abiosoft (approved):

abiosoft:
Thanks :)

PR #512, opened 2022-12-14
dns: support for custom hosts

FabianPonce:
This PR adds support for passing host resolver information to Lima, allowing the internal DNS resolver to resolve any custom names you wish. This is particularly useful for services that do not use `getaddrinfo`, ie. specific nginx, envoy configurations.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -37,6 +37,7 @@ Run 'colima template' to set the default configurations or 'colima start --edit'
 		"  colima start --cpu 4 --memory 8 --disk 100\n" +
 		"  colima start --arch aarch64\n" +
 		"  colima start --dns 1.1.1.1 --dns 8.8.8.8\n" +
+		"  colima start --dns-host example.com=1.2.3.4\n" +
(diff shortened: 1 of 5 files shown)


abiosoft:
Hi, thanks for this.

Rather than bringing the entire `hostResolver` config into Colima, I would recommend supporting config for only the custom hosts. 

It can easily be an array flag e.g. `--custom-host=host.docker.internal=host.lima.internal`
And a config section for `customHosts` that is a `map[string]string`.

We would also need to keep `host.docker.internal` as default/base value and only append the user specified values.

FabianPonce:
I agree, I think that would be a much cleaner solution in general. I'll make some changes and re-request review.

FabianPonce:
@abiosoft I felt that the custom hosts functionality, being related to DNS, should live somewhere near the resolver settings, so I moved this under the network section. I'm happy to break it out into the originally suggested location if you still think that's more suitable. 

I also renamed the DNS Resolvers list in the config struct to `DNSResolvers` to make that more clear in the code. I left the naming in the yaml alone as `network.dns`, though, to keep backwards compatibility for people using that.

abiosoft (approved):

PR #528, opened 2022-12-20
vm: revert to qemu as default

abiosoft:


rfay:
I appreciated this, and think it was the right move. I would have expected to roll back to sshfs as default mount type at the same time, as 9p also has some issues, including performance, but also it has hangs on many file changes, for example `ddev composer install` normally hangs somewhere in the process. (It's just a `composer install` on a mount inside the container)

abiosoft:
@rfay thanks for the heads up.

PR #545, opened 2022-12-27
vm: bring back sshfs, workaround virtiofs

### LINKED ISSUE #544: Consider going back to sshfs as default mount type, at least for QEMU; 9p performance is awful
### Description

We had hopes for 9p performance, and still should, but right now it's terrible. 

I just did a set of performance comparisons of different Colima and Docker Desktop characteristics in https://ddev.com/ddev-local/ddev-docker-desktop-and-colima-benchmarking-updated-dec-2022/ and found that the QEMU/9p setup wasn't even worth waiting for; it's as bad as Docker Desktop alone used to be. 

I think we should consider sshfs as the default still until things shake out. However, my experience with Virtiofs was not bad at all, except for the serious performance problem described in https://github.com/drud/ddev/issues/4466#issuecomment-1361261185 and https://github.com/Code-Hex/vz/pull/117, which should make it into new versions of Lima in the near future.

abiosoft:
Fixes #544

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -95,7 +95,12 @@ const (
 	defaultMemory            = 2
 	defaultDisk              = 60
 	defaultKubernetesVersion = kubernetes.DefaultVersion
-	defaultNetworkDriver     = "gvproxy"
+
+	defaultNetworkDriver = "gvproxy"
+
+	defaultVMType        = "qemu"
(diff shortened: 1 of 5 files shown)


abiosoft:
the failed job is not an issue, it's due to the slow emulation of aarch64.

PR #551, opened 2022-12-30
chore: update module dependencies

rfay:
I was trying to do a PR to add validation to config, but was unable to use debugger (arm64, goland). 

I ran into 
* https://github.com/golang/go/issues/54291#issuecomment-1368037141

The fix seems to be to upgrade:
* `go get -u gvisor.dev/gvisor/runsc@go`
and for good measure, upgrade the rest of modules
* `go get -u ./...`

which is what is done here.

References:
* [gvisor using go get](https://github.com/google/gvisor#using-go-get)
* https://github.com/google/gvisor/issues/8237
* https://github.com/google/gvisor/issues/8149


abiosoft:
The amount of delete lines from the `go.sum` file is quite surprising.

abiosoft (approved):

abiosoft:
> I was trying to do a PR to add validation to config, but was unable to use debugger (arm64, goland).

Wow, thanks :)

Issue #277, opened 2022-05-10
`--add-host=host.docker.internal:host-gateway` doesn't work like it does on Linux


Atemu:
For repro, start something on the host that listens on a certain port like `python3 -m http.server`.

Then run

```
docker run -it --rm --add-host=host.docker.internal:host-gateway nixery.dev/shell/pingtcp pingtcp host.docker.internal 8000
```

On a Linux machine, this successfully pings port 8000. With colima, it's unable to handshake because host.docker.internal now resolves to the Linux VM where the python http server isn't running.

This flag (and its equivalent docker-compose declaration) is often used in order to give Linux users' containers access to the host address because host.docker.internal doesn't work there by default.

abiosoft:
> This flag (and its equivalent docker-compose declaration) is often used in order to give Linux users' containers access to the host address because host.docker.internal doesn't work there by default.

But host.docker.internal works with docker containers on Colima, or does it not?

Atemu:
Yes, that works by default but not when adding this flag, which is often done by docker-compose files for compatibility with Linux.

paymog:
looks like this issue may be stale. I have a docker compose file with that host mapping for compatibility across linux and mac. However, when trying out my compose file with colima, everything worked flawlessly.

```
 ❯❯❯ colima version
colima version 0.4.6
git commit: 10377f3a20c2b0f7196ad5944264b69f048a3d40

runtime: docker
arch: aarch64
client: v20.10.20
server: v20.10.18
```

rfay:
host.docker.internal is always 192.168.5.2 on colima/lima.

efgpinto:
I think I'm facing the same issue.

While on colima:

```bash
$ colima version
colima version HEAD-030393c
git commit: 030393ccc5ab74153809ed00dbe965325d3cd7af

$ python3 -m http.server
Serving HTTP on :: port 8000 (http://[::]:8000/) ...
```

Cannot do the handshake:
```bash
$ docker run -it --rm --add-host=host.docker.internal:host-gateway nixery.dev/shell/pingtcp pingtcp host.docker.internal 8000
(...)
Digest: sha256:d35346995775beb2f3b1cf0f7c4411ca493359a791cbf1dedd3d1cc2cf4f5367
Status: Downloaded newer image for nixery.dev/shell/pingtcp:latest
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
[...]

PR #574, opened 2023-01-06
docker: set host-gateway-ip to loopback interface

artygus:
This PR intends to improve compatibility with docker desktop to simplify migration to colima.

https://docs.docker.com/engine/reference/commandline/dockerd/

> --host-gateway-ip ip 
> IP address that the special 'host-gateway' string in --add-host resolves to. Defaults to the IP address of the default bridge

Docker desktop sets this setting to `192.168.65.2` which is loopback address for vpnkit.

Fixes https://github.com/abiosoft/colima/issues/277


abiosoft (approved):

abiosoft:
Thanks :)

Issue #398, opened 2022-08-16
Disk migration / upgrade disk size manually


Morgy93:
### Description

Hello,

is there some way of disk migration?
The only thing I can find about disk ist the following:
>NOTE: disk size cannot be changed after the VM is created.

While this is totally understandable, is there some way of migration?
Like copy foo from the VM to the host via bar, delete VM, create new one and copy foo to VM at baz.

SuperFola:
Since this is a Linux VM you have access to things like ssh and scp, one could copy files out using those, and then back in.

Morgy93:
> Since this is a Linux VM you have access to things like ssh and scp, one could copy files out using those, and then back in.

Do you have any idea about all the necessary files and folders?
A very simple overview of the inner structure would help closing this issue.

SuperFola:
It depends on what you want to save, if it's the docker volumes you can list them with a `fdisk -l`. Otherwise I wouldn't care too much, and just delete the VM to create a new one.

abiosoft:
If it's primarily with regards to docker (or OCI) images, you can `docker save` to a .tar file and `docker load` after a reset.

Morgy93:
> If it's primarily with regards to docker (or OCI) images, you can `docker save` to a .tar file and `docker load` after a reset.

Well, the main request for this that with Docker Desktop you can adjust the VM size without any data loss.
Currently that's not possible with Colima so I asked for some sort of documentation for manual upgrading "without any data loss". 
But I assume that the images are the least required data, more important are the volumes for example.

artygus:
hey @Morgy93 you should be able to use `qemu-img` for the task, e.g. `qemu-img resize ~/.lima/colima/diffdisk +10G`, the change needs vm restart

rfay:
Oh my goodness, if that works could you add a PR for the docs @artygus ? We all thought you had to recreate. Of course only works for qemu.

jandubois:
> Oh my goodness, if that works[...]

This requires lima 0.14+ (https://github.com/lima-vm/lima/pull/1126).

artygus:
@jandubois thanks for the notes! one problem noticed is that lima reports the initial size in `limactl list`, are there plans to fix it?

jandubois:
> one problem noticed is that lima reports the initial size in `limactl list`, are there plans to fix it?

I'm not aware; please file an issue on the Lima repo!

[...]

PR #582, opened 2023-01-12
Add disk adjustment instructions for qemu

artygus:
This PR adds instructions for disk adjustment for qemu. The changes made by `qemu-img` will be picked up by lima on next boot https://github.com/lima-vm/lima/blob/f26b412/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh#L16-L26.

Not sure aboute wording, feel free to modify it 😅 

Addresses issue mentioned in https://github.com/abiosoft/colima/issues/398.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Under vm-type qemu disk size can be adjusted using the `qemu-img` utility
 qemu-img resize ~/.lima/colima/diffdisk +10G
 ```
 
-Make sure to restart colima.
+Disk size will be resized on the VM next boot if it was created with lima 0.14+.
 
 **NOTE**: disk size cannot be changed with vm-type vz once the VM is created.
 
--- a/README.md
+++ b/README.md
@@ -111,7 +111,15 @@ The VM can be customized either by passing additional flags to `colima start`.
 e.g. `--cpu`, `--memory`, `--disk`, `--runtime`.
 Or by editing the config file with `colima start --edit`.
 
-**NOTE**: disk size cannot be changed after the VM is created.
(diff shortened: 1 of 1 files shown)


lulor:
I'm pretty sure you can resize the disk using `qemu-img` even on VZ machines, although you're advised to pass the additional parameter `-f raw` (not sure about the format when using QEMU).
I think that a potential problem here is that Lima/Colima's info about the disk size size may become incoherent with the actual size of the image. Maybe it would be better if Colima checked the existence of `diffdisk` and, in case it already exists, report its actual size when queried?

artygus:
@lulor yeah, I've noticed that, colima takes info about the disk size from lima, which also reports it incorrectly, not sure where it should be fixed. can't check it for vz atm, can you try that?

lulor:
I've tried to resize the disk of a VZ machine (`qemu-img resize -f raw .lima/colima/diffdisk`) and it looks like it's working fine.

Since Colima generates the Lima config at runtime, maybe it would be better to check the existence of `diffdisk` and, it case it exists, store its size as the output of `ls -lh .lima/colima/diffdisk | awk '{print $5}'`?

lulor:
Oh I see..
Sure then, `qemu-img info .lima/colima/diffdisk | awk '/virtual size/ {print $3}'` should do it

abiosoft:
Since this is now supported by Lima, I am working on implementing this directly in Colima.

i.e. simply changing disk size via `--disk` flag or config should be all that is required.

artygus:
@abiosoft thanks for the update! that's grand news! I'm closing this PR.

rfay:
Huge improvement, thanks both!

PR #602, opened 2023-01-31
cli: add restart command

rfay:
I've long wanted a `colima restart` command and so here is one :)


abiosoft (approved):

abiosoft:
> I've long wanted a `colima restart` command and so here is one :)

same here actually :)

Issue #500, opened 2022-12-07
Colima volume sync issue on M1 Max


invaderb:
### Description

I have 2 apple m1 computers I'm working on, one is a m1 pro and the other is a m1 max.  I have a git repo with a docker file and a docker compose file to set up a few containers for local development.  I initially set everything up on the m1 pro and got everything working however I need to now switch over to my m1 max and when I went to set everything up on the m1 max same running the same commands. it appears that the the volume isn't getting synced like the m1 pro is doing. any help would be greatly appreciated.

### Version

Colima Version:
colima version v0.4.6-19-gf959232
git commit: f959232a2322e4b61217ecba0b92ef4618783cbf

runtime: docker
arch: aarch64
client: v20.10.21
server: v20.10.20

Lima Version: installed latest with brew

Qemu Version: what version is in the latest version from brew

### Operating System

- [ ] macOS Intel
- [X] macOS M1
- [ ] Linux

### Reproduction Steps

1.
2.
3.

### Expected behaviour

volume sync should happen

invaderb:
Ok so I was reading through the FAQ and saw that I can run both docker desktop and colima together, so I'm able to run part of my app on docker desktop which resolves my volume sync issue and I can run my mssql container with colima which is a work around but not an ideal solution.

abiosoft:
@invaderb by getting synced, are you referring to live-reload?

invaderb:
@abiosoft It might also be called live reload I just know it as volume sync in the docker compose file

    volumes:
      - .:/var/www/html
      
Where it sets the current directory to the destination folder in the container.

abiosoft:
@invaderb the `sshfs` mount sometimes can have some delays. If you do not mind resetting your setup, can you try the `9p` mount?

```sh
colima delete # reset
colima start --mount-type 9p
```

Next version of Colima would have the option of utilising native macOS virtualization which should resolve this for most users.

invaderb:
@abiosoft would it also be the same on the vz-support branch which is what I'm currently on for the rosetta 2 support?

abiosoft:
@invaderb oh wait, I didn't check your version properly and assumed you were on stable. You should not have any volume related issues.

Can you kindly share the output of `colima status`?

invaderb:
@abiosoft 

INFO[0000] colima is running using macOS Virtualization.Framework 
INFO[0000] arch: aarch64                                
INFO[0000] runtime: docker                              
INFO[0000] mountType: sshfs                             
[...]

PR #606, opened 2023-02-02
[Meta] Update README with Ventura mount-type reset

prdanelli:
I was running into an issue with MacOS Ventura where by all directory volumes - where I link a directory on my host machine to one on the container - created with `docker-compose` are empty. I've tested the same scripts on MacOS 12.4 and 12.6 and they work as expected giving me the same directory contents on the container as on the host, so it seems v13 changed some permission. 

The `docker-compose.yml` file:

```
version: "3"

services:
  bash:
    image: ubuntu:latest
    stdin_open: true
    tty: true
    volumes:
      - ./:/app
    command: "/bin/bash"
```

So this should be creating a directory on the container called `/app` and linking that to the host directory the compose file is in.

But when I start the container:

```
❯ docker-compose up --build
[+] Running 1/0
 ⠿ Container ruby-docker-bash-1  Created  
Attaching to ruby-docker-bash-1
```

And login, the `/app` directory is empty:

```
❯ docker exec -it ruby-docker-bash-1 /bin/bash
root@9644de175d48:/# cd app/
root@9644de175d48:/app# ls -la
total 4
drwxr-xr-x 2 root root   40 Feb  1 09:59 .
drwxr-xr-x 1 root root 4096 Feb  1 09:39 ..
```

The `total 4` is really weird here as there are 4 files supposed to be there, but not accessible:

```
root@9644de175d48:/app# cat Gemfile
cat: Gemfile: No such file or directory
```

This is the directory contents on the host are:

```
❯ ls -la
.rw-r--r-- 2.7k paul  1 Feb 09:31 Dockerfile
.rw-r--r-- 3.9k paul  1 Feb 09:32 Gemfile
.rw-r--r--  27k paul  1 Feb 09:32 Gemfile.lock
.rw-r--r--  149 paul  1 Feb 10:00 docker-compose.yml
```

Then I found this: https://github.com/abiosoft/colima/issues/500#issuecomment-1343103477

```
colima delete # reset
colima start --mount-type 9p
```

I’ve been through the site, the readme, (there is no Wiki) and no where is the mount types documented, I guess it might be a Docker thing.

But [I did find this](https://github.com/abiosoft/colima/blob/51406badb45697da1d4cc2d0cb8225f7e2a3e55e/config/configmanager/config.go#L68-L70)…

```	
validMountTypes := map[string]bool{"9p": true, "sshfs": true}
if util.MacOS13OrNewer() {
	validMountTypes["virtiofs"] = true
}
```

I’m in MacOS 13, so it seems like there are issues with `virtiofs` and not in the older `9p` mount type, so i'd like to suggest adding a line to the README that might help others that run into this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,15 @@ Or by editing the config file with `colima start --edit`.
   colima start --arch aarch64 --vm-type=vz --vz-rosetta
   ```
 
+- set the mount-type to `9p` (defaults to `virtiofs` on MacOS Ventura)
+
+  ```
+  colima delete # reset
+  colima start --mount-type 9p
+  ```
+
+  If you are experiencing issues with volume mounting directories in MacOS Ventura (13.x) this may help, by resetting the mount type back to its default.
+
 ## Project Goal
 
 To provide container runtimes on macOS with minimal setup.


abiosoft:
Yeah, `virtiofs` should only accompany native the macOS virtualization. Something is off it's being set to virtiofs when Qemu is being used.

What version of Colima are you on?

prdanelli:
Hello,

This is the Colima version I have installed: 

```
❯ colima version
colima version 0.5.2
git commit: 6b5b6fe0540e708f0c9d6e8919fab292c671fc72

runtime: docker
arch: aarch64
client: v20.10.22
server: v20.10.20
```

abiosoft:
This is actually resolved and would be part of the next release. i.e. Qemu would use sshfs/9p mounts and macOS virtualization will use virtiofs.

prdanelli:
@abiosoft thank you for clarifying. Do you know when the next release is going to be live?

abiosoft:
> @abiosoft thank you for clarifying. Do you know when the next release is going to be live?

About to :)

PR #608, opened 2023-02-05
vm: respect `$SHELL` in `colima ssh` when `layer=true`

satoqz:
This PR replaces the hardcoded `bash` used when logging into the ubuntu layer with `$SHELL`.
This is useful in case the user sets their login shell to a different shell, e.g. zsh or fish.

It also goes in line with the way that lima handles ssh login itself:
https://github.com/lima-vm/lima/blob/6ae40d3a25a6ffa006bbc04d0e8433357e756e8a/cmd/limactl/shell.go#L124-L129


abiosoft (approved):

abiosoft:
Thanks :)

PR #615, opened 2023-02-09
Update README.md

mindhells:
see this thread https://stackoverflow.com/questions/72557053/why-does-colima-failed-to-find-docker-daemon


abiosoft:
Something similar is already in the FAQ

mindhells:
> Something similar is already in the FAQ

missed it, ty

PR #620, opened 2023-02-14
Update FAQ.md

ktomana:
I couldn't start Colima on my M1 macbook and setting `arch: aarch64` in the default config helped. Maybe this information will help others :)


onatm:
`colima start --kubernetes` worked completely fine on my Macbook Pro M1 Ventura 13.1.

Could you please verify what do you see on your terminal when you run `arch` @ktomana?

I had a similar issue with other tools and it turned out `tmux` was running under x86 emulation.

abiosoft:
This should not be required as it defaults to the native OS architecture.

PR #658, opened 2023-03-20
vm: fix startup delay for inotify

abiosoft:


abiosoft:
The failed build can be disregarded

Issue #261, opened 2022-04-20
NodeJS hot reload issue


tcastelly:
Hello,

I switched from Docker Desktop to Colima. I noticed an issue with "watcher" on bind volume. I created a repository: https://github.com/shenron/colima-hot-reload-issue

The container is not able to see when a file has been updated. But if I touch the file from the container it works. I guess it's linked to the volume.

---

I'm using the Macbook Pro M1 Max with:
```
colima version 0.3.4
git commit: 5a4a70481ca8d1e794677f22524e3c1b79a9b4ae

runtime: docker
arch: aarch64
client: v20.10.14
server: v20.10.11
```

```
limactl version 0.9.2
```

ggoodman:
The way colima binds host directories into the guest VM means that certain filesystem events are not observed / generated.

You can work around this by using `--legacy-watch` in `nodemon` that will fall back to polling instead of the [fsevents](https://npm.im/fsevents) integration.

luiz290788:
I'm facing the same issue with `ts-node-dev`. Is this an issue that will eventually be solved by colima or we will never receive the same events when using colima?

abiosoft:
>  Is this an issue that will eventually be solved by colima or we will never receive the same events when using colima?

I am working on a workaround taking a similar approach to https://github.com/abiosoft/colima/issues/261#issuecomment-1116466858 while leveraging inotify.  

I paused due to some issues with the Go library at the time, but I will revisit it.

matthewdickinson:
> > Is this an issue that will eventually be solved by colima or we will never receive the same events when using colima?
> 
> I am working on a workaround taking a similar approach to [#261 (comment)](https://github.com/abiosoft/colima/issues/261#issuecomment-1116466858) while leveraging inotify.
> 
> I paused due to some issues with the Go library at the time, but I will revisit it.

It would be great to have this feature completed. Is there a work-in-progress branch that I could possibly help continue/test?

djirik:
Hi, is there any progress on that? If you need some help, I may push our development team to help you. We need this so our frontend team could work using colima. Docker for mac is not an option for us)

abiosoft:
@djirik support for **virtiofs** mounts is imminent and would be part of the next release. 

It has been merged into master upstream in Lima and currently being integrated and tested in Colima, the early results have been quite impressive.

Virtiofs is same tech used by Docker Desktop for the fast volume speeds.

djirik:
@abiosoft thank you very much

[...]

PR #669, opened 2023-03-27
Notify improvements

henrik242:
A couple of changes that improve notify events (at least for me)

Ref #261


abiosoft:
Thanks, that was quick :).

However, I already did something similar in https://github.com/abiosoft/colima/pull/668. You can try out the branch

henrik242 (commented):
This works pretty reliably now. Great work!

PR #671, opened 2023-03-27
vm: mount: improve error messages and reduce noise

abiosoft:


daemon/process/inotify/events.go
```diff
@@ -92,15 +91,18 @@ func (f *inotifyProcess) handleEvents(ctx context.Context, watcher dirWatcher) e
...
-				log.Trace(err)
+				log.Trace(fmt.Errorf("cannot stat '%s': %w", ev.path, err))
 				continue
 			}
 
-			log.Infof("refreshing mtime for %s ", ev.path)
-			if err := f.guest.RunQuiet("/bin/chmod", ev.Mode(), ev.path); err != nil {
-				log.Error(err)
+			log.Infof("syncing inotify event for %s ", ev.path)
+			if err := f.guest.RunQuiet("sudo", "/bin/chmod", ev.Mode(), ev.path); err != nil {
```
henrik242, line 104: @abiosoft We don't always have sudo rights in the container

abiosoft: That's not running in the container. It runs on the VM directly and there is always sudo access.

PR #709, opened 2023-05-11
net: add option to pass static IP for col0 interface via env variable

aelsnz:
One of the issues were are seeing many report with MacOS 13 Ventura is when you use the `--network-address` option is that bootpd is not giving out a DHCP address.

In many cases you get error similar to below - for more details see https://github.com/abiosoft/colima/issues/683

```
time="2023-05-11T12:37:27+12:00" level=fatal msg="error starting kubernetes: error running [lima kubectl cluster-info], output: \"The connection to the server localhost:8080 was refused - did you specify the right host or port?\", err: \"exit status 1\""
```

This change basically allow you to specify an environment variable COLIMA_IP with an IP address in subnet 192.168.106.0/24 
When the `ifaces.sh` script is executed it will check if it detects this environment variable it will set the static ip address for the `col0` interface which is used when the `--network-address` flag is specified.

This is not a perfect solution, but it gives us an option to get past this problem.

With this change you can run start using: 

```
colima start -c 1 -d 10 -m 2 --network-address --env COLIMA_IP=192.168.106.201
```

And  now you get the IP address assigned and the environment is usable.

### DIFF
--- a/embedded/network/ifaces.sh
+++ b/embedded/network/ifaces.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env sh
+# set -x
 
 FILE=/etc/network/interfaces
 BACKUP="$FILE.bak"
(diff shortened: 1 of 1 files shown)


abiosoft (approved):

abiosoft:
Thanks for this :)

Issue #724, opened 2023-05-24
Unable to start colima, when path of home folder contains BLANK


bigguyong:
### Description

Unable to start colima, seems caused by the home folder's path, it contains a BLANK. For privacy reason, I replace the path name of home folder to "AAAAA BB"

{"level":"debug","msg":"Creating iso file /Users/**AAAAA BB**/.lima/colima/cidata.iso","time":"2023-05-24T17:31:03+08:00"}
{"level":"debug","msg":"Using /var/folders/rk/35js3v794kq79srwrx20mj800000gp/T/diskfs_iso3638406399 as workspace","time":"2023-05-24T17:31:03+08:00"}
{"level":"debug","msg":"Failed to detect CPU features. Assuming that AES acceleration is available on this Apple silicon.","time":"2023-05-24T17:31:03+08:00"}
{"level":"debug","msg":"OpenSSH version 9.0.1 detected","time":"2023-05-24T17:31:04+08:00"}
{"level":"debug","msg":"AES accelerator seems available, prioritizing aes128-gcm@openssh.com and aes256-gcm@openssh.com","time":"2023-05-24T17:31:04+08:00"}
{"level":"warning","msg":"Specifying args ([BB/.colima/_wrapper/3a9197e1ca3cd2da076da2b473d7a7eb118e2cca/bin/qemu-system-aarch64]) via $QEMU_SYSTEM_AARCH64 is supported only for debugging!","time":"2023-05-24T17:31:04+08:00"}
{"level":"fatal","msg":"exec: \"**/Users/AAAAA**\": stat **/Users/AAAAA**: no such file or directory","time":"2023-05-24T17:31:04+08:00"}

### Version

Colima Version: 0.5.5
Lima Version:0.15.1
Qemu Version:8.0.0

### Operating System

- [ ] macOS Intel <= 12 (Monterrey)
- [ ] macOS Intel >= 13 (Ventura)
- [ ] macOS M1 <= 12 (Monterrey)
- [X] macOS M1 >= 13 (Ventura)
- [ ] Linux

### Output of `colima status`

FATA[0000] colima is not running

### Reproduction Steps

1.
2.
3.

watersb:
I just completed a patch for Colima to use the [XDG_Base_Dirs specification](https://wiki.archlinux.org/title/XDG_Base_Directory) if things like XDG_CONFIG_HOME are defined in the environment.

[...]

PR #728, opened 2023-06-01
Fix #483 - Use XDG Base Dir or platform dirs for config files

### LINKED ISSUE #483: Feature request: change "home" directory location, and/or support XDG Base Directory spec
### Description

It would be really nice if we could change the location of the Colima "home" directory to something other than `~/.colima`.

For example, I tend to create directories in `~/.local/opt` for tools like this, so it would be nice if I could set `export COLIMA_HOME=$HOME/.local/opt/colima` in my system configuration.

Alternatively, it would be nice if Colima supported the [XDG Base Directory Specification](https://xdgbasedirectoryspecification.com/). The config files would be located in `XDG_CONFIG_HOME`, the daemon socket could be located in `XDG_STATE_HOME`, etc.

I think many people nowadays like to keep as few "dotfiles" in their home directory as possible, and this would be a nice quality of life addition for those users.

watersb:
If the environment variable XDG_CONFIG_HOME is set, the Colima configuration should be (shell syntax) `"${XDG_CONFIG_HOME}/colima`.

Likewise, if XDG_CACHE_HOME is set, then the cache should be in `"${XDG_CACHE_HOME}/colima`.

If these variables are not set, use the Go standard lib `os.UserConfigDir()` for the configuration. _On macOS, that will be `$HOME/Library/Application Support/colima`; on Linux it defaults to `$HOME/.config/colima`_.

This patch addresses #483, and possibly helps #724.

### DIFF
--- a/config/dirs.go
+++ b/config/dirs.go
@@ -42,12 +42,19 @@ func (r *requiredDir) Dir() string {
 var (
 	configBaseDir = requiredDir{
 		dir: func() (string, error) {
-			dir := os.Getenv("XDG_CONFIG_HOME")
+			dir, err := os.UserHomeDir()
+			dir = filepath.Join(dir, ".colima")
+			_, err = os.Stat(dir)
+			if err == nil {
+				return dir, nil
+			}
+			// else
+			dir = os.Getenv("XDG_CONFIG_HOME")
 			if dir != "" {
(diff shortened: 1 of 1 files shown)


watersb:
Oops -- if the current existing default location `${HOME}/.colima` exists. we should use that.

PR #736, opened 2023-06-09
chore: use XDG base dir or platform dirs for config files. Fixes #483.

### LINKED ISSUE #483: Feature request: change "home" directory location, and/or support XDG Base Directory spec

watersb:
This patch add some options for the location of user configuration directories.

As before, use `$HOME/.colima` if it already exists.

Otherwise, if the environment variable `XDG_CONFIG_HOME` is set, the Colima configuration directory should be (shell syntax) `"${XDG_CONFIG_HOME}"/colima`.

Likewise, if `XDG_CACHE_HOME` is set, then the cache should be in `"${XDG_CACHE_HOME}"/colima`.

If these variables are _not_ set, use the Go standard lib `os.UserConfigDir()` for the configuration:

-  `macOS  ` $HOME/Library/Application Support/colima
-  `Linux `  $HOME/.config/colima
- `Windows` %AppData%\colima

and uses `os.UserCacheDir()` for the cache:

-  `macOS  ` $HOME/Library/Caches/colima
-  `Linux `  $HOME/.cache/colima
- `Windows` %LocalAppData%\colima

This patch addresses https://github.com/abiosoft/colima/issues/483, and possibly helps https://github.com/abiosoft/colima/issues/724.

### DIFF
--- a/config/dirs.go
+++ b/config/dirs.go
@@ -42,12 +42,19 @@ func (r *requiredDir) Dir() string {
 var (
 	configBaseDir = requiredDir{
 		dir: func() (string, error) {
-			dir := os.Getenv("XDG_CONFIG_HOME")
+			dir, err := os.UserHomeDir()
+			dir = filepath.Join(dir, ".colima")
+			_, err = os.Stat(dir)
+			if err == nil {
+				return dir, nil
+			}
+			// else
+			dir = os.Getenv("XDG_CONFIG_HOME")
 			if dir != "" {
 				return filepath.Join(dir, "colima"), nil
 			}
 			// else
-			dir, err := os.UserConfigDir()
+			dir, err = os.UserConfigDir()
 			if err != nil {
 				return "", err
 			}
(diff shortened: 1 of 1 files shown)


abiosoft:
Thanks for this :)

abiosoft (approved):

PR #789, opened 2023-08-29
feat(start): allow to start colima in the foreground

regulskimichal:
Currently, Colima does not provide a working brew service. To make it possible, it should implement foreground mode. I'm pretty new to Golang but I think awaiting an interruption after the startup should be enough.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -3,8 +3,10 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/abiosoft/colima/cli"
@@ -49,6 +51,9 @@ Run 'colima template' to set the default configurations or 'colima start --edit'
 			if app.Active() {
 				log.Warnln("already running, ignoring")
 				return nil
+			}
+			if startCmdArgs.Flags.Foreground {
+				defer awaitForInterruption()
 			}
 			return app.Start(conf)
 		}
@@ -126,6 +131,7 @@ var startCmdArgs struct {
 		Editor                  string
 		ActivateRuntime         bool
 		DNSHosts                []string
+		Foreground              bool
 	}
 }
 
@@ -145,6 +151,7 @@ func init() {
 	startCmd.Flags().IntVarP(&startCmdArgs.Memory, "memory", "m", defaultMemory, "memory in GiB")
 	startCmd.Flags().IntVarP(&startCmdArgs.Disk, "disk", "d", defaultDisk, "disk size in GiB")
 	startCmd.Flags().StringVarP(&startCmdArgs.Arch, "arch", "a", defaultArch, "architecture (aarch64, x86_64)")
+	startCmd.Flags().BoolVarP(&startCmdArgs.Flags.Foreground, "foreground", "f", false, "Keep colima in the foreground")
 
 	// network
 	if util.MacOS() {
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -3,6 +3,7 @@
 - [FAQs](#faqs)
   - [How does Colima compare to Lima?](#how-does-colima-compare-to-lima)
   - [Are M1 macs supported?](#are-m1-macs-supported)
+  - [Does Colima support autostart](#does)
```
adam-at-mobomo, line 6: The link should be: `#does-colima-support-autostart`

cmd/start.go
```diff
@@ -45,6 +48,10 @@ Run 'colima template' to set the default configurations or 'colima start --edit'
 		app := newApp()
 		conf := startCmdArgs.Config
 
+		if startCmdArgs.Flags.Foreground {
+			defer awaitForInterruption(&app)
```
abiosoft, line 52: This should be called only after a successful start. i.e. `app.Start(conf)` does not return an error.

abiosoft:
Thnaks for this :)

regulskimichal:
I wasn't aware of other modes of colima. It looks like some tests/cleanup may require additional changes

cmd/start.go
regulskimichal: Updated

regulskimichal:
I've found out that integration tests fail because of: https://github.com/abiosoft/colima/issues/796
My changes should not affect other modes of Colima

regulskimichal:
@abiosoft Any chances for merge 🥺 ?

abiosoft:
> @abiosoft Any chances for merge 🥺 ?

Yeah, soon.

cmd/start.go
```diff
@@ -443,3 +449,23 @@ func editConfigFile() error {
...
 	return configmanager.SaveFromFile(tmpFile)
 }
+
+func start(app *app.App, conf *config.Config) error {
+	startResult := (*app).Start(*conf)
+	if startResult == nil && startCmdArgs.Flags.Foreground {
+		defer awaitForInterruption(app)
+	}
+	return startResult
+}
```
abiosoft, line 459:
```suggestion
func start(app app.App, conf config.Config) error {
	if err := app.Start(conf); err != nil {
		return err
	}
	if startCmdArgs.Flags.Foreground {
		awaitForInterruption(app)
	}
	return nil
}
```

cmd/start.go
```diff
@@ -443,3 +449,23 @@ func editConfigFile() error {
...
+	signalChannel := make(chan os.Signal, 1)
+	signal.Notify(signalChannel, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+	log.Println("Keeping Colima in the foreground, press ctrl+c to exit...")
+	sig := <-signalChannel
+	log.Debug("Interrupted by:", sig)
+
+	if err := (*app).Stop(false); err != nil {
+		_ = fmt.Errorf("error stopping :%w", err)
+	}
+}
```
abiosoft, line 471:
```suggestion
func awaitForInterruption(app app.App) {
	signalChannel := make(chan os.Signal, 1)
	signal.Notify(signalChannel, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
	log.Println("keeping Colima in the foreground, press ctrl+c to exit...")
	sig := <-signalChannel
	log.Infof("interrupted by: %v", sig)

	if err := app.Stop(false); err != nil {
		log.Errorf("error stopping :%v", err)
	}
}
```

cmd/start.go
regulskimichal: Is it ok to pass by value?

cmd/start.go
regulskimichal: 👍

cmd/start.go
abiosoft: It has no effect because it is an interface, which points to an actual value.

abiosoft (approved):

PR #828, opened 2023-10-07
chore: remove redundant `len` check

Juneezee:
From the Go specification (https://go.dev/ref/spec#For_range):

> "3. If the map is nil, the number of iterations is 0."

`len` returns 0 if the map is nil (https://pkg.go.dev/builtin#len). Therefore, checking `len(v) > 0` around a loop is unnecessary.


abiosoft (approved):

abiosoft:
Thanks :)

PR #858, opened 2023-11-13
Minor update to issue template

rfay:
I had just a little trouble navigating the issue template today, so trying to improve slightly. I think `sonoma` is getting increasingly important, so put that in there.


abiosoft (approved):

abiosoft:
Thanks as usual :)

PR #864, opened 2023-11-14
Allow dev version of lima

terev:
The latest release of colima made it impossible to use a dev version of lima. This change allows the user to use a dev version but prints a warning.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -127,11 +127,7 @@ func LimaVersionSupported() error {
 	}
 
 	if parts[0] == "HEAD" {
-		version := parts[0]
-		if len(parts) > 1 {
-			version = parts[1]
-		}
-		logrus.Warnf("to avoid compatibility issues, ensure lima development version (%s) in use is not lower than %s", version, limaVersion)
+		logrus.Warnf("to avoid compatibility issues, ensure lima development version (%s) in use is not lower than %s", values.Version, limaVersion)
 		return nil
 	}
 
--- a/core/core.go
+++ b/core/core.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/sirupsen/logrus"
 	"strings"
 
 	"github.com/abiosoft/colima/cli"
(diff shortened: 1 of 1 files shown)


abiosoft:
Thanks for creating this.

As a side note, if you did a `git pull --tags` prior to running `make` for Lima, you'd get a version that would pass the version check, because the most recent tag name would be used.

```
$ limactl --version
limactl version 0.18.0-88-g79cfe42.m
```

terev:
Oh I see, I installed the version of lima using homebrew so it must use a different version. Should the homebrew recipe be changed instead?

core/core.go
```diff
@@ -120,8 +120,18 @@ func LimaVersionSupported() error {
...
+	if len(parts) > 0 {
+		values.Version = parts[0]
+	}
+
+	if parts[0] == "HEAD" {
+		version := parts[0]
+		if len(parts) > 1 {
+			version = parts[1]
+		}
+		fmt.Printf("Lima Version Warning: Using a dev version of lima (%s). If the used version isn't a child of %s you may experience issues.\n", version, limaVersion)
```
AkihiroSuda, line 133: logrus.Warnf

AkihiroSuda, line 133: What is "child" ?

terev: Any commit that is a child of v0.18.0 is what I meant.

core/core.go
```diff
@@ -120,8 +121,18 @@ func LimaVersionSupported() error {
...
+	if len(parts) > 0 {
+		values.Version = parts[0]
+	}
+
+	if parts[0] == "HEAD" {
+		version := parts[0]
+		if len(parts) > 1 {
+			version = parts[1]
+		}
+		logrus.Warnf("Using a dev version of lima (%s). If the used version isn't a child of %s you may experience issues.\n", version, limaVersion)
```
abiosoft, line 134:
```suggestion
		logrus.Warnf("to avoid compatibility issues, ensure lima development version (%s) in use  is not lower than %s, version, limaVersion)
```

terev: Made this change. Just fixed some formatting in your suggestion.

AkihiroSuda:
I suggest to squash commits

abiosoft (dismissed):

core/core.go
```diff
@@ -120,8 +121,18 @@ func LimaVersionSupported() error {
...
+		values.Version = parts[0]
+	}
+
+	if parts[0] == "HEAD" {
+		version := parts[0]
+		if len(parts) > 1 {
+			version = parts[1]
+		}
+		logrus.Warnf("to avoid compatibility issues, ensure lima development version (%s) in use is not lower than %s", version, limaVersion)
+		return nil
```
abiosoft, line 135: I just realised there is actually no need to extract any version details anymore.

```suggestion
	if parts[0] == "HEAD" {
		logrus.Warnf("to avoid compatibility issues, ensure lima development version (%s) in use is not lower than %s", values.Version, limaVersion)
		return nil
```

abiosoft (approved):

AkihiroSuda:
@abiosoft Could you consider making a release with this?

PR #885, opened 2023-11-18
vm: purge packages instead of just removing them

LaszloGombos:
Without purging, packages not fully removed (e.g. dpkg -l will still show the packages).


rfay (approved):
I noticed this when experimenting with the official packages as well, purge is probably better here.

abiosoft (approved):

PR #887, opened 2023-11-18
vm: do not require optional system utilities

LaszloGombos:
"htop", "vim", "inetutils-ping", "dnsutils" are optional dependencies - only required for developers working on colima.

I'd like to get some feedback on possibly moving these packages out of the main codebase into the configuration file. This would enable developers removing these packages without patching colima core source code.

Thanks !

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -123,12 +123,12 @@ jobs:
 
       - name: Delay
         run: sleep 10
-      
+
       - name: Validate Docker
         run: docker ps && docker info
 
       - name: Validate DNS
-        run: colima ssh -- nslookup host.docker.internal
+        run: colima ssh -- sh -c "sudo apt-get update -y -qq && sudo apt-get install -qq dnsutils && nslookup host.docker.internal"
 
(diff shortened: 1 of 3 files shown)


abiosoft:
I do agree that they can be moved outside of the main codebase.

The integration tests are failing because `nslookup` is assumed to exist for docker local domain name validation. 
You would need modify the tests to manually install `dnsutils` in the VM.

abiosoft:
The goal is to eventually have a custom generated Ubuntu image with only the necessary packages bundled.

abiosoft:
> Should I completely remove them, move them to `embedded/defaults/colima.yaml` or perhaps to `embedded/defaults/template.yaml`?

I think they can be completely removed and example in the `provision` section of the config should suffice.

abiosoft (approved):

abiosoft:
> Yes, make sense. is the idea to build up an image with [debootsratp](https://help.ubuntu.com/community/DebootstrapChroot) or simply purge packages from the current cloud image ?

I do not know which one is easier. Actually we can create a separate issue to track that.

PR #890, opened 2023-11-19
vm: do not install recommended packages

LaszloGombos:
See https://www.debian.org/doc/manuals/debian-faq/pkg-basics.en.html#depends for the definition of Recommended packages.

Not installing Recommended packages makes the generation more controllable and predictable.

Currently, in practice as an example docker-ce will no longer install docker-ce-rootless-extras .


abiosoft (approved):

abiosoft:
Thanks :)

PR #892, opened 2023-11-19
docker: allow specifying version and channel to install

crazy-max:
I have a use case where I would need to specify the docker version/channel to install. This is related to a GitHub Action I maintain https://github.com/crazy-max/ghaction-setup-docker and specially https://github.com/crazy-max/ghaction-setup-docker/issues/47.

With this change, user can set `DOCKER_INSTALL_VERSION` and `DOCKER_INSTALL_CHANNEL` to install a specific version. This also returns the right `pkgFiles` based on the selected channel.

### DIFF
--- a/environment/vm/lima/deb/docker.go
+++ b/environment/vm/lima/deb/docker.go
@@ -2,6 +2,7 @@ package deb
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/abiosoft/colima/environment"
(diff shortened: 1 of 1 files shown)


environment/vm/lima/deb/docker.go
```diff
@@ -30,8 +31,12 @@ func (d *Docker) PreInstall() error {
 
 // Install implements URISource.
 func (d *Docker) Install() error {
+	args := []string{"docker", "--channel", dockerChannel()}
+	if v := os.Getenv("DOCKER_INSTALL_VERSION"); v != "" {
+		args = append(args, "--version", v)
+	}
 	return d.Guest.Run("sh", "-c",
-		`curl -fsSL https://get.docker.com -o /tmp/get-docker.sh && sudo sh /tmp/get-docker.sh`,
+		`curl -fsSL https://get.docker.com | sh -s `+strings.Join(args, " "),
```
abiosoft, line 39: I think you can still pass the args to the written file i.e.
```
... sudo sh /tmp/get-docker.sh` + strings.Join(args, " ")
```

crazy-max: I found it better to directly pipe to the shell so the script is executed immediately without being stored locally but if you prefer to keep writing the script before exec let me know.

crazy-max: Updated

abiosoft (approved):

PR #893, opened 2023-11-19
vm: support additional disks

ssakuh:
adds support for a new parameter in colima.yaml that will create and mount additional disks via limactl


abiosoft:
`limactl disk create` command is not idempotent. i.e. there needs to be a check if the disk does not exist before running the command.

abiosoft:
Hi, thanks for your efforts so far.

Can you kindly limit this PR to only support additional disks? Other fixes can happen in a separate PR.

stephg:
I would love to see support for additional disks merged. Besides separating all the cosmetic fixes in this PR, does this need some more work?

abiosoft:
This PR is would need to be closed, unfortunately. It is too behind (evident from the conflicts) and contains irrelevant changes.

The ground work has actually been done to support additional disks https://github.com/abiosoft/colima/commit/eac8ef6b6247810e941c7ed980814714923bd495. I am not sure how soon I will get to it but I have it planned.

PR #907, opened 2023-11-26
deps: add opt to skip caching dependencies

crazy-max:
When implementing #892, I haven't realized that cached dependencies take precedence over the standard installation which handles specifying version and channel for docker installation.

This PR adds an env var `COLIMA_SKIP_CACHE_DEPS` so it can do a direct install and skip caching dependencies. This also makes sense when used in CI environment like GitHub Actions where runners are ephemeral.

Let me know if an env var and its name makes sense.

### DIFF
--- a/environment/vm/lima/dependencies.go
+++ b/environment/vm/lima/dependencies.go
@@ -79,6 +79,15 @@ func (l *limaVM) installDependencies(log *logrus.Entry, conf config.Config) erro
 			log.Warn(fmt.Errorf("preinstall check failed for %s: %w", src.Name(), err))
 		}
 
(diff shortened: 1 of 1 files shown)


abiosoft:
Actually, for a better user experience, I am working on bundling dependencies in the underlying cloud image and thereby the whole dependency caching thing would be removed.

I cannot give an ETA but it is not far off.

At the same time, I can see your use case and the benefit of being able to specify Docker version (similar to k3s version). I am exploring how best to accomodate it.

As a worst case scenario, users would be able to `colima ssh` and manually uninstall the bundled Docker and install any desired Docker version.

crazy-max:
Thanks sounds good! Feel free to close this PR

PR #918, opened 2023-12-05
support ssh local port customization

elvuel:
Random ports are annoying sometimes.


abiosoft:
FYI an ssh config is auto generated/updated on startup/shutdown.

If the sole purpose of having a fixed port is the convenience to use ssh, you can rely on the generated config.

```sh
ssh colima # always works
```

PR #924, opened 2023-12-10
vm: default to vz for m3 devices

abiosoft:


AkihiroSuda:
The M3 issue is expected to be fixed in:
- https://github.com/lima-vm/lima/pull/2061

PR #946, opened 2023-12-26
k3s: Update default template version

azdanov:
Bumping k3s template version to avoid confusion as in https://github.com/abiosoft/colima/discussions/945


abiosoft (approved):

abiosoft:
Thanks :)

PR #978, opened 2024-02-06
k3s: write kubeconfig to KUBECONFIG defined file

Sumoa:
Currently when starting a vm + k8s cluster the kubeconfig is always written to "~/.kube/config" which is the default path. By defining the env variable "KUBECONFIG" it is possible to override the default path. This does currently not apply to colima.
As it is possible to configure multiple locations with "KUBECONFIG" I opted to always use the first configured one as the location to save the config.

### DIFF
--- a/environment/container/kubernetes/kubeconfig.go
+++ b/environment/container/kubernetes/kubeconfig.go
@@ -3,7 +3,6 @@ package kubernetes
 import (
 	"context"
 	"fmt"
-	"os"
 	"path/filepath"
 	"strings"
 	"time"
--- a/environment/container/kubernetes/kubeconfig.go
+++ b/environment/container/kubernetes/kubeconfig.go
@@ -42,6 +42,10 @@ func (c kubernetesRuntime) provisionKubeconfig(ctx context.Context) error {
 	})
 
 	kubeconfFile := filepath.Join(hostKubeDir, "config")
+	envKubeConfFile := c.host.Env("KUBECONFIG")
+	if envKubeConfFile != "" {
+		kubeconfFile = strings.Split(envKubeConfFile, ":")[0]
+	}
 	tmpkubeconfFile := filepath.Join(hostKubeDir, "."+profile, "colima-temp")
 
 	// manipulate in VM and save to host
--- a/environment/container/kubernetes/kubeconfig.go
+++ b/environment/container/kubernetes/kubeconfig.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -44,7 +45,7 @@ func (c kubernetesRuntime) provisionKubeconfig(ctx context.Context) error {
 	kubeconfFile := filepath.Join(hostKubeDir, "config")
 	envKubeConfFile := c.host.Env("KUBECONFIG")
 	if envKubeConfFile != "" {
-		kubeconfFile = strings.Split(envKubeConfFile, ":")[0]
+		kubeconfFile = strings.Split(envKubeConfFile, string(os.PathListSeparator))[0]
 	}
 	tmpkubeconfFile := filepath.Join(hostKubeDir, "."+profile, "colima-temp")
 
--- a/environment/container/kubernetes/kubeconfig.go
+++ b/environment/container/kubernetes/kubeconfig.go
@@ -45,7 +45,7 @@ func (c kubernetesRuntime) provisionKubeconfig(ctx context.Context) error {
 	kubeconfFile := filepath.Join(hostKubeDir, "config")
 	envKubeConfFile := c.host.Env("KUBECONFIG")
 	if envKubeConfFile != "" {
-		kubeconfFile = strings.Split(envKubeConfFile, string(os.PathListSeparator))[0]
+		filepath.SplitList(envKubeConfFile)[0]
 	}
 	tmpkubeconfFile := filepath.Join(hostKubeDir, "."+profile, "colima-temp")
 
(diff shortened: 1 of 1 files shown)


environment/container/kubernetes/kubeconfig.go
```diff
@@ -42,6 +42,10 @@ func (c kubernetesRuntime) provisionKubeconfig(ctx context.Context) error {
 	})
 
 	kubeconfFile := filepath.Join(hostKubeDir, "config")
+	envKubeConfFile := c.host.Env("KUBECONFIG")
+	if envKubeConfFile != "" {
+		kubeconfFile = strings.Split(envKubeConfFile, ":")[0]
```
adam-at-mobomo, line 47: To make this Windows-friendly, you could use `os.PathListSeparator` instead of `":"`

Sumoa: While I agree that using `os.PathListSeparator` is better, I would really like to know why `Colima - container runtimes on **macOS (and Linux)** with minimal setup.` would need to be more windows-friendly? It does not run on Windows as it is a macOS and Linux **ONLY** project.

abiosoft:
@Sumoa can you kindly rebase? I fixed the broken CI.

environment/container/kubernetes/kubeconfig.go
abiosoft: You both have valid points. Considering that Colima is still far from being able to properly support windows, we do not have to worry about it.

However, using `os.PathListSeparator` is pretty much effortless.

environment/container/kubernetes/kubeconfig.go
```diff
@@ -42,6 +43,10 @@ func (c kubernetesRuntime) provisionKubeconfig(ctx context.Context) error {
 	})
 
 	kubeconfFile := filepath.Join(hostKubeDir, "config")
+	envKubeConfFile := c.host.Env("KUBECONFIG")
+	if envKubeConfFile != "" {
+		kubeconfFile = strings.Split(envKubeConfFile, string(os.PathListSeparator))[0]
```
abiosoft, line 48: Actually there is an in-built function for this `filepath.SplitList` https://pkg.go.dev/path/filepath@go1.22.2#SplitList.

salzig:
i would prefer a command to show me the kubeconfig so i can redirect to a file as i like, in addition to a config option to NOT alter my kubeconfig. I need to work with quite many clusters, which are also changing quite often. So i started to use separate readonly files (~/.kube/config.d/) and load them all as `KUBECONFIG` (`export KUBECONFIG=$(ls -1 ~/.kube/config.d/*.yaml | tr "\n" ":")`). Altering ~/.kube/config isn't that useful for me. 

Respecting the KUBECONFIG would at least allow me to workaround the problem. Thanks @Sumoa.

hedefalk:
I'd love to get this too! I'm always using $KUBECONFIG using direnv's .envrc files. That's the best way (only?) to have different default kube contexts per folder.

I've tried to forward the k3s arg `--write-kubeconfig` via `--k3s-arg` but I think that is all in vain as I guess colima does this another way because I need it written out to the host machine and what I forward to --write-kubeconfig would need to be the mount point within the vm, right? Maybe a a combination of that at mounting the KUBECONFIG into the vm would be an option but I'm just feeling too thick right now :)

Currently I need to manually copy over the sections from ~/.kube/config into my contextual KUBECONFIG. I also get this error message in the end of a start because it can't find the config it just thought it wrote out (since it writes it out to a different place than what kubectl sees):

`FATA[0025] error starting kubernetes: error at 'updating config': error running [kubectl config use-context colima-kb], output: "error: no context exists with the name: \"colima-kb\"", err: "exit status 1"`

hedefalk:
While looking at the source code I realized I could get away with a script, so I added this bash script to my dotfiles/stow repo that kinda does all I want:

```bash
#!/bin/bash

# Check if COLIMA_PROFILE is provided
if [ -z "$1" ]; then
    echo "Usage: $0 <COLIMA_PROFILE>"
    exit 1
fi

COLIMA_PROFILE=$1
COLIMA_KUBE_CONTEXT=colima-$COLIMA_PROFILE
TEMP_COLIMA_KUBECONFIG=$(mktemp)
TEMP_MERGED_KUBECONFIG=$(mktemp)
colima -p $COLIMA_PROFILE ssh -- kubectl config view --raw  | sed "s/default/$COLIMA_KUBE_CONTEXT/g" > $TEMP_COLIMA_KUBECONFIG
EXISTING_KUBECONFIG=$KUBECONFIG
KUBECONFIG=$KUBECONFIG:$TEMP_COLIMA_KUBECONFIG
kubectl config view --raw > $TEMP_MERGED_KUBECONFIG
cp $TEMP_MERGED_KUBECONFIG $EXISTING_KUBECONFIG
KUBECONFIG=$EXISTING_KUBECONFIG
rm $TEMP_COLIMA_KUBECONFIG $TEMP_MERGED_KUBECONFIG
kubectl config set-context $COLIMA_KUBE_CONTEXT
```


With that I can keep working wih direnv and have a .envrc file in each "contextual" dev folder doing:

    export KUBECONFIG=$(pwd)/kubeconfig

Then I can first setup say a local stage cluster with k3sup (that is already honoring $KUBECONFIG):

    k3sup install --skip-install --host pi1 --user pi --merge --context k3s 

and then add colima k3s to this:

```
colima start -p kb
export-colima-profile kb

kubectl config get-contexts

CURRENT   NAME        CLUSTER     AUTHINFO    NAMESPACE
*         colima-kb   colima-kb   colima-kb
          k3s         k3s         k3s
```

environment/container/kubernetes/kubeconfig.go
```diff
@@ -42,6 +42,10 @@ func (c kubernetesRuntime) provisionKubeconfig(ctx context.Context) error {
 	})
 
 	kubeconfFile := filepath.Join(hostKubeDir, "config")
+	envKubeConfFile := c.host.Env("KUBECONFIG")
+	if envKubeConfFile != "" {
+		filepath.SplitList(envKubeConfFile)[0]
```
abiosoft, line 47:
```suggestion
		kubeconfFile = filepath.SplitList(envKubeConfFile)[0]
```

Issue #1051, opened 2024-06-26
APFS Volume not correctly mounting in docker container


nnathan:
### Description

Volume mount working with default system volume:

```
c2:code $ df -h .
Filesystem      Size    Used   Avail Capacity iused ifree %iused  Mounted on
/dev/disk3s5   926Gi   636Gi   269Gi    71%    2.0M  2.8G    0%   /System/Volumes/Data
c2:code $ docker run -it --mount type=bind,src=$PWD,dst=/model ubuntu
root@9daa7afabc1d:/# cd /model/
root@9daa7afabc1d:/model# ls
DocExample.pdf          frr                             musl                         spiped
  .
  .
  .
root@9daa7afabc1d:/# mount | grep -i model
mount0 on /model type virtiofs (rw,relatime)
```

Volume mount **not working** using a separate volume mount (case sensitive volume):

```
c2:landingnet $ df -h .
Filesystem      Size    Used   Avail Capacity iused ifree %iused  Mounted on
/dev/disk3s7   926Gi   1.5Gi   269Gi     1%     21k  2.8G    0%   /Volumes/s
c2:landingnet $ docker run -it --mount type=bind,src=$PWD,dst=/model ubuntu
root@8b715762cba0:/# cd /model/
root@8b715762cba0:/model# ls
root@8b715762cba0:/model# mount | grep -i model
/dev/vda1 on /model type ext4 (rw,relatime,discard,errors=remount-ro,commit=30)
```

### Version

```
c2:colima $ colima version && limactl --version && qemu-img --version
colima version 0.6.9
git commit: c3a31ed05f5fab8b2cdbae835198e8fb1717fd0f

runtime: docker
arch: aarch64
client: v26.1.3
server: v26.1.1
limactl version 0.22.0
qemu-img version 9.0.1
Copyright (c) 2003-2024 Fabrice Bellard and the QEMU Project developers
```

### Operating System

- [ ] macOS Intel <= 13 (Ventura)
- [ ] macOS Intel >= 14 (Sonoma)
- [ ] Apple Silicon <= 13 (Ventura)
- [X] Apple Silicon >= 14 (Sonoma)
- [ ] Linux

### Output of `colima status`

```
c2:colima $ colima status
INFO[0000] colima is running using macOS Virtualization.Framework
INFO[0000] arch: aarch64
INFO[0000] runtime: docker
INFO[0000] mountType: virtiofs
[...]

PR #1052, opened 2024-06-27
docs: update faq regarding docker bind mount issue

nnathan:
Although most people won't run into the issue as they're probably mapping volumes from their `/Users/<username>` directory, I think this information would be handy for anyone running into the issue of docker bind mounts showing up empty on the container. This was the issue I faced in https://github.com/abiosoft/colima/issues/1051 and it took me 1-2 days to solve it.


abiosoft (approved):

abiosoft:
Thanks :)

Issue #987, opened 2024-02-19
Colima hangs on macOS, cannot stop


erickrf:
### Description

Colima hanged on my macOS installation. The VMs I try to start simply get stuck. If I run `colima status`, it just hangs there forever.

`colima stop` returns `INFO[0000] stopping colima` and then also hangs.

`colima start` just returns `WARN[0000] already running, ignoring`. 

The only way out was to manually kill the process `qemu-system_x86_64` and restart colima, but then it froze again after running a VM.

### Version

```
colima version 0.6.8
git commit: 9b0809d0ed9ad3ff1e57c405f27324e6298ca04f
limactl version 0.20.1
qemu-img version 8.2.1
Copyright (c) 2003-2023 Fabrice Bellard and the QEMU Project developers
```

While it was in the stuck behavior, only the first two lines are displayed, then it hangs.

### Operating System

- [X] macOS Intel <= 13 (Ventura)
- [ ] macOS Intel >= 14 (Sonoma)
- [ ] Apple Silicon <= 13 (Ventura)
- [ ] Apple Silicon >= 14 (Sonoma)
- [ ] Linux

### Output of `colima status`

(hangs)

### Reproduction Steps

It happens when running some specific VM.

rommelfreddy:
can confirm. i have the same issue on Sonoma 14.2 / Colima 0.6.7

Using it with DDEV - projects working properly, and after some times the containers does not respond can colima is not stoppable.

After killing all colima/lima processes, and starting colima and the containers again, the containers does response but very slowly. 

Also having performance issues with xdebug - but this may also be a problem on the DDEV or xdebug.

ngamsomset:
same here. it stuck while trying to download something.
<img width="562" alt="image" src="https://github.com/abiosoft/colima/assets/60306190/f61b2188-103d-4fc8-9e45-85e2d1e72eba">

dsmooot:
same issue as @ngamsomset.. 
M1 MBP Sonoma 14.2.1

thejan2009:
I can confirm the above, MBP M1 Pro, Sonoma 14.3.1, Colima 0.6.8. After starting the disk expansion, the whole system becomes slow. Activity monitor reports around 15MB/sec is written to the disk. That keeps happening even after I stop colima. Only a reboot stops this behavior.

megalucio:
Can confirm as well MBP M1 Pro Ventura 13.6.3, hangs and no way to fix other than reboot

[...]

Issue #985, opened 2024-02-13
colima start does not honor default config, but then overwrites it


pangdaxing23:
### Description

The `vz` VM type does not work on my M3 Mac, so I must use `qemu`. My config at `~/.config/default/colima.yaml` has `vmType: qemu` so that when I run `colima start` I should use the `qemu` VM type. Instead it goes with the `vz` type, and overwrites the config file as well. It also fails for this same reason when running `brew services start colima`, and indeed overwrites the config too. I am able to run `colima start --edit`, see that it has already overwritten the `vmType`, edit it to `qemu`, then exit and it will work, but a simple `colima start` fails. The only way I have ever gotten it to work in brew services temporarily is to run `colima delete` then reinstall colima, but it breaks again after some time, and I am not sure what triggers it, possibly a reboot.

### Version

colima version 0.6.8
git commit: 9b0809d0ed9ad3ff1e57c405f27324e6298ca04f
limactl version 0.20.1
qemu-img version 8.2.1

### Operating System

- [ ] macOS Intel <= 13 (Ventura)
- [ ] macOS Intel >= 14 (Sonoma)
- [ ] Apple Silicon <= 13 (Ventura)
- [X] Apple Silicon >= 14 (Sonoma)
- [ ] Linux

### Output of `colima status`

If I run `colima start`, then `colima status`:
```
FATA[0000] colima is not running
```
If I run `colima start --edit`, and edit it to `qemu`, then run `colima status`:
```
INFO[0000] colima is running using QEMU
INFO[0000] arch: aarch64
INFO[0000] runtime: docker
INFO[0000] mountType: virtiofs
INFO[0000] socket: unix:///Users/MYUSER/.colima/default/docker.sock
```

### Reproduction Steps

1. Set `vmType: qemu` in `~/.colima/default/colima.yaml`.
2. Run `colima start`.
3. Note failure and check config to see it over.written with `vmType: vz`

### Expected behaviour

`colima start` should run the profile in the config and not overwrite the config.

### Additional context

`colima start` gives the following:
```INFO[0000] starting colima
INFO[0000] runtime: docker
INFO[0000] starting ...                                  context=vm
> Using the existing instance "colima"
> errors inspecting instance: [vz driver is running but host agent is not]
FATA[0000] error starting vm: error at 'starting': exit status 1
```

philippe-lammerts-remote:
I have exactly the same problem on the same setup

mostafa6765:
[...]

PR #1058, opened 2024-07-04
core: revert to using qemu by default on M3 macs

bradleyayers:
This removes the special case logic for M3 macs that use the vz driver instead of qemu. A lot of people are having trouble with vz on M3 macs:

- https://github.com/abiosoft/colima/issues/987#issuecomment-1999938930
- https://github.com/abiosoft/colima/pull/1045#issue-2363549238
- https://github.com/abiosoft/colima/issues/985#issue-2132482350
- https://github.com/abiosoft/colima/issues/985#issuecomment-2024729107
- https://github.com/abiosoft/colima/issues/985#issuecomment-1987058790

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -241,12 +241,6 @@ func mountsFromFlag(mounts []string) []config.Mount {
 func setDefaults(cmd *cobra.Command) {
 	if startCmdArgs.VMType == "" {
 		startCmdArgs.VMType = defaultVMType
-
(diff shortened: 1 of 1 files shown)


abiosoft:
Thanks.

It was meant to be a stop-gap solution at the time when Qemu wasn't working on m3 macs. Now it makes sense to remove it.

abiosoft (approved):

Issue #1060, opened 2024-07-04
Provision scripts are ignored when VM is just created


Noksa:
### Description

Provision scripts are ignored when we start VM first time (create it)

### Version

```shell
colima version 0.6.9
git commit: c3a31ed05f5fab8b2cdbae835198e8fb1717fd0f

runtime: docker
arch: aarch64
client: v27.0.3
server: v26.1.1
limactl version 0.22.0
qemu-img version 9.0.1
Copyright (c) 2003-2024 Fabrice Bellard and the QEMU Project developers
```

### Operating System

- [ ] macOS Intel <= 13 (Ventura)
- [ ] macOS Intel >= 14 (Sonoma)
- [ ] Apple Silicon <= 13 (Ventura)
- [X] Apple Silicon >= 14 (Sonoma)
- [ ] Linux

### Output of `colima status`

```shell
INFO[0000] colima is running using macOS Virtualization.Framework
INFO[0000] arch: aarch64
INFO[0000] runtime: docker
INFO[0000] mountType: virtiofs
INFO[0000] socket: unix:///Users/alexandr/.colima/default/docker.sock
INFO[0000] kubernetes: enabled
````

### Reproduction Steps

1. Add `provision` block in template as follows:

```yaml
provision:
  - mode: system
    script: |
      cat <<EOF > /etc/systemd/system/docker.service.d/docker-buildkit-env.conf
      [Service]
      Environment="BUILDKIT_STEP_LOG_MAX_SIZE=20971520"
      Environment="BUILDKIT_STEP_LOG_MAX_SPEED=1048576"
      EOF
      systemctl daemon-reload
      systemctl restart docker
```

2. Clean install VM

```shell
colima start
```

3. Check that file is not present: `colima ssh -- ls /etc/systemd/system/docker.service.d`

4. Stop and start again
```shell
colima stop && colima start
```

5. Now file is present

### Expected behaviour

Provision scripts should be considered when we do `start` first time

### Additional context

As a workaround we can use `colima start --edit` but just save template without changes. In this case, scripts will be run

davidolrik:
I too have this problem and the workaround does not work for me.

My custom provision section are never executed.

```
colima version 0.7.5
git commit: 1588c066b9ab9dae8205ef265929c7eb43dca473

runtime: docker
arch: aarch64
client: v27.2.0
server: v27.1.1
```

blast-hardcheese:
[...]

Issue #834, opened 2023-10-20
Insecure Registry not being set


Taylor150:
### Description

I have use colima start --edit to add my insecure registry:
`docker:
  insecure-registries:
    - 192.168.172.27:6969
    - repo.local:6969
    - localhost:6969
    - 127.0.0.1:6969`

 however when i do a docker login, it looks like the insecure registry is not being recognised:
`docker login http://192.168.178.27:6969
Username: admin
Password:
Error response from daemon: Get "https://192.168.178.27:6969/v2/": http: server gave HTTP response to HTTPS client`

Any suggestions?

### Version

Colima Version: 0.5.6
Lima Version: 0.18.0
Qemu Version: 8.1.2

### Operating System

- [ ] macOS Intel <= 12 (Monterrey)
- [ ] macOS Intel >= 13 (Ventura)
- [ ] macOS M1 <= 12 (Monterrey)
- [X] macOS M1 >= 13 (Ventura)
- [ ] Linux

### Output of `colima status`

colima is running using macOS Virtualization.Framework
arch: aarch64
runtime: docker
mountType: virtiofs
address:
socket: unix:///Users/laurencetaylor/.colima/default/docker.sock

### Reproduction Steps

1. I have a local docker registry
2. I have set this registry in the insecure-registries within colima config
3. I have tried to docker login to the registry/docker pull
4. I receive and error which infers that the insecure registry is not recognised 

### Expected behaviour

Be able to login to my insecure registry

abiosoft:
Can you kindly share the output of the following.

```sh
# display the generated docker config
colima ssh -- cat /etc/docker/daemon.json; echo
```

Taylor150:
> Can you kindly share the output of the following?
colima ssh -- cat /etc/docker/daemon.json; echo
{
  "exec-opts": [
    "native.cgroupdriver=cgroupfs"
  ],
  "features": {
    "buildkit": true
  },
  "host-gateway-ip": "192.168.5.2"
}

Taylor150:
Not sure if related but i also notice these when colima is starting:
`WARN[0054] unable to disable qemu x86_84 emulation: exit status 1`
`stat: can't stat '/proc/sys/fs/binfmt_misc/qemu-x86_64': No such file or directory`

klit79:
Me too
It was not working even if colima restart.
I had to run this
`colima ssh sudo service docker restart`

amirkhonov:
```
colima ssh sudo systemctl daemon-reload
[...]

PR #1097, opened 2024-08-06
fix(docker): apply configuration on first boot

milas:
The initialization sequence is:
 (1) VM boot
 (2) Write daemon configuration

When the VM boot, the Docker systemd service is started, so the new configuration is not picked up until the next VM boot.

To fix this, reload the systemd daemon config (to pick up the systemd service file written by `addHostGateway`) and then restart the Docker systemd service, which will use the new systemd unit file as well as any customizations to `daemon.json`.

### Related Issues
* #834
* #1060

(My specific issue was that my `daemon.json` customizations in the `docker` section of `colima.yaml` were not being applied on first boot.)

### DIFF
--- a/environment/container/docker/daemon.go
+++ b/environment/container/docker/daemon.go
@@ -77,6 +77,16 @@ func (d dockerRuntime) addHostGateway(conf map[string]any) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


abiosoft:
Thanks :)

PR #1113, opened 2024-08-19
chore: add second confirmation for container data loss

jramosf:
We have observed some users not realizing container data is lost when doing delete. This change does not modify scripted behaviour, as -f / force toggles maintains same logic.

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -28,6 +28,10 @@ If you simply want to reset the Kubernetes cluster, run 'colima kubernetes reset
 			if !y {
 				return nil
 			}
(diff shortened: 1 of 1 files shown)


abiosoft:
Thanks :) I actually got bitten by this recently as well :D 

I am planning to work on a more rubust approach that would separate container data entirely from the instance. i.e. deleting the instance would not delete the containers. You would need to specify a flag (`--volumes` maybe) to `colima delete` to get the containers deleted.

That would solve many problems and would allow you to upgrade (and reset) Colima versions without the fear of losing data.

abiosoft (approved):

PR #1196, opened 2024-11-15
cli: add --json flag to status command

ecrousseau:
Adds the option of JSON output to the status command, which would be useful for me, and possibly others.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -317,7 +317,6 @@ func (c colimaApp) getStatus() (status statusInfo, err error) {
 		return status, err
 	}
 
-	var status statusInfo
 	status.DisplayName = config.CurrentProfile().DisplayName
 	status.Driver = "QEMU"
 	conf, _ := configmanager.LoadInstance()
--- a/app/app.go
+++ b/app/app.go
@@ -30,7 +30,7 @@ type App interface {
 	Stop(force bool) error
 	Delete() error
 	SSH(args ...string) error
-	Status(extended bool) error
+	Status(extended bool, jsonOutput bool) error
 	Version() error
 	Runtime() (string, error)
 	Update() error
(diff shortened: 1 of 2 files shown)


cmd/status.go
```diff
@@ -16,6 +17,9 @@ var statusCmd = &cobra.Command{
 	Long:  `Show the status of Colima`,
 	Args:  cobra.MaximumNArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if statusCmdArgs.output == "json" {
```
abiosoft, line 20:
```suggestion
		if statusCmdArgs.json {
```

abiosoft:
Thanks for this :)

ecrousseau:
I've pushed an update that avoids changing the interface, as you suggested. My thinking on using `--output json` rather than just `--json` was that it leaves room to add other output formats in the future. Easy to change it to a boolean flag if you'd prefer, though.

abiosoft:
> My thinking on using --output json rather than just --json was that it leaves room to add other output formats in the future.

Considering that `--json` is already used for the `colima list` command, I would prefer sticking to `--json` for consistency.

ecrousseau:
Ah I did not realise there was an existing pattern - completely agree it should stay consistent!

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
-	log.Println("arch:", c.guest.Arch())
-	log.Println("runtime:", currentRuntime)
-	if conf.MountType != "" {
-		log.Println("mountType:", conf.MountType)
+	status.Arch = fmt.Sprintf("%v", c.guest.Arch())
+	status.Runtime = currentRuntime
+	status.MountType = conf.MountType
+	ipAddress := limautil.IPAddress(config.CurrentProfile().ID)
+	if ipAddress != "127.0.0.1" {
+		status.IPAddress = limautil.IPAddress(config.CurrentProfile().ID)
```
abiosoft, line 332: There is no need to do the lookup again.

```suggestion
		status.IPAddress = ipAddress
```

abiosoft (dismissed):

app/app.go
```diff
@@ -30,7 +30,7 @@ type App interface {
 	Stop(force bool) error
 	Delete() error
 	SSH(args ...string) error
-	Status(extended bool) error
+	Status(extended bool, jsonOutput bool) error
```
abiosoft, line 33:
```suggestion
	Status(extended bool, json bool) error
```

abiosoft (dismissed):

cmd/status.go
```diff
@@ -16,12 +17,16 @@ var statusCmd = &cobra.Command{
 	Long:  `Show the status of Colima`,
 	Args:  cobra.MaximumNArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return newApp().Status(statusCmdArgs.extended)
+		if statusCmdArgs.json {
+			return newApp().Status(statusCmdArgs.extended, true)
+		}
+		return newApp().Status(statusCmdArgs.extended, false)
```
abiosoft, line 23:
```suggestion
		return newApp().Status(statusCmdArgs.extended, statusCmdArgs.json)
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
-
-	// ip address
-	if ipAddress := limautil.IPAddress(config.CurrentProfile().ID); ipAddress != "127.0.0.1" {
-		log.Println("address:", ipAddress)
-	}
-
-	// docker socket
 	if currentRuntime == docker.Name {
-		log.Println("socket:", "unix://"+docker.HostSocketFile())
+		status.DockerSocket = fmt.Sprintf("unix://%s", docker.HostSocketFile())
```
abiosoft, line 335:
```suggestion
		status.DockerSocket = "unix://" + docker.HostSocketFile()
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
+		}
+
+		// ip address
+		if status.IPAddress != "" {
+			log.Println("address:", status.IPAddress)
+		}
+
+		// docker socket
+		if status.DockerSocket != "" {
+			log.Println("socket:", "unix://"+status.DockerSocket)
```
abiosoft, line 373:
```suggestion
			log.Println("socket:", status.DockerSocket)
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
 	return guest.SSH(workDir, args...)
 }
 
-func (c colimaApp) Status(extended bool) error {
+type status struct {
```
abiosoft, line 295:
```suggestion
type statusInfo struct {
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
 	}
 
 	currentRuntime, err := c.currentRuntime(ctx)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	driver := "QEMU"
+	var status status
```
abiosoft, line 320:
```suggestion
	var status statusInfo
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
-		driver = conf.DriverLabel()
+		status.Driver = conf.DriverLabel()
 	}
-
-	log.Println(config.CurrentProfile().DisplayName, "is running using", driver)
-	log.Println("arch:", c.guest.Arch())
-	log.Println("runtime:", currentRuntime)
-	if conf.MountType != "" {
-		log.Println("mountType:", conf.MountType)
+	status.Arch = fmt.Sprintf("%v", c.guest.Arch())
```
abiosoft, line 327:
```suggestion
	status.Arch = string(c.guest.Arch())
```

abiosoft (dismissed):

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
+	MountType    string `json:"mount_type"`
+	IPAddress    string `json:"ip_address"`
+	DockerSocket string `json:"docker_socket"`
+	Kubernetes   bool   `json:"kubernetes"`
+	CPU          int    `json:"cpu"`
+	Memory       int64  `json:"memory"`
+	Disk         int64  `json:"disk"`
+}
+
+func (c colimaApp) getStatus() (*status, error) {
```
abiosoft, line 309:
```suggestion
func (c colimaApp) getStatus() (status statusInfo, err error) {
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
 	ctx := context.Background()
 	if !c.guest.Running(ctx) {
-		return fmt.Errorf("%s is not running", config.CurrentProfile().DisplayName)
+		return nil, fmt.Errorf("%s is not running", config.CurrentProfile().DisplayName)
```
abiosoft, line 312:
```suggestion
		return status, fmt.Errorf("%s is not running", config.CurrentProfile().DisplayName)
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
 	}
 
 	currentRuntime, err := c.currentRuntime(ctx)
 	if err != nil {
-		return err
+		return nil, err
```
abiosoft, line 317:
```suggestion
		return status, err
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
 	if k, err := c.Kubernetes(); err == nil && k.Running(ctx) {
-		log.Println("kubernetes: enabled")
+		status.Kubernetes = true
+	}
+	if inst, err := limautil.Instance(); err == nil {
+		status.CPU = inst.CPU
+		status.Memory = inst.Memory
+		status.Disk = inst.Disk
 	}
+	return &status, nil
```
abiosoft, line 345:
```suggestion
	return status, nil
```

app/app.go
```diff
@@ -292,54 +292,105 @@ func (c colimaApp) SSH(args ...string) error {
...
-		}
+func (c colimaApp) Status(extended bool, jsonOutput bool) error {
+	status, err := c.getStatus()
+	if err != nil {
+		return err
 	}
 
+	if jsonOutput {
+		b, _ := json.Marshal(status)
+		fmt.Println(string(b))
```
abiosoft, line 356:
```suggestion
		if err := json.NewEncoder(os.Stdout).Encode(status); err != nil {
			return fmt.Errorf("error encoding status as json: %w", err)
		}
```

app/app.go
```diff
@@ -292,54 +292,106 @@ func (c colimaApp) SSH(args ...string) error {
...
 
 	currentRuntime, err := c.currentRuntime(ctx)
 	if err != nil {
-		return err
+		return status, err
 	}
 
-	driver := "QEMU"
+	var status statusInfo
+	status.DisplayName = config.CurrentProfile().DisplayName
```
abiosoft, line 321:
```suggestion
	status.DisplayName = config.CurrentProfile().DisplayName
```

abiosoft (approved):

abiosoft:
Once again, thanks.

PR #1227, opened 2024-12-21
Add supports for k0s

chenhunghan:


abiosoft:
Thanks for your efforts, really appreciated.

I wish that you opened a discussion before going all the way to implement this.

I think this is a bit of a stretch. The decision to go with k3s was not to offer varieties for Kubernetes, rather it was the easiest choice to integrate and so far has sufficed as the Kubernetes on Colima.

Currently, adding k0s into the mix is out of scope for Colima. What is more inline is to swap k3s with k0s, but I am yet to see any user demands in that regard or notable benefits to consider it.

Once again, thanks.

PR #1235, opened 2024-12-31
chore: update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Abiola Ibrahim
+Copyright (c) 2025 Abiola Ibrahim
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


abiosoft:
Thanks :)

abiosoft (approved):

windyakin:
The year of the copyright is to indicate the year of first publication, not to update it.

https://courier.unesco.org/en/articles/universal-copyright-convention

> It prescribes that the formalities required by the national law of a contracting state shall be considered to be satisfied if all the copies of a work originating in another contracting state carry the symbol ©, accompanied by the name of the copyright owner and **the year of first publication**.

abiosoft:
> The year of the copyright is to indicate the year of first publication, not to update it.

Thanks for sharing.

If that is indeed the case, this PR should be reverted @JasonnnW3000 .

Issue #1239, opened 2025-01-05
Log error if exits when fetching Colima config


olamilekan000:
### Description

Actual errors are not displayed when they occur during config reading process.

Issue was created based on this [ticket](https://github.com/abiosoft/colima/issues/1169).

PR #1240, opened 2025-01-05
chore: log possible error while reading config file

olamilekan000:
```
change logs actual error when trying to read Colima config file
```

Issue can be found [here](https://github.com/abiosoft/colima/issues/1239).

### DIFF
--- a/environment/vm/lima/config.go
+++ b/environment/vm/lima/config.go
@@ -15,7 +15,7 @@ func (l limaVM) getConf() map[string]string {
 	obj := map[string]string{}
 	b, err := l.Read(configFile)
 	if err != nil {
-		log.Tracef("failed to read config file %v", err)
+		log.Trace(fmt.Errorf("error reading config file: %w", err))
 
 		return obj
 	}
--- a/environment/vm/lima/config.go
+++ b/environment/vm/lima/config.go
@@ -1,6 +1,7 @@
 package lima
 
 import (
+	"context"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


environment/vm/lima/config.go
```diff
@@ -9,9 +10,13 @@ import (
 const configFile = "/etc/colima/colima.json"
 
 func (l limaVM) getConf() map[string]string {
+	log := l.Logger(context.Background())
+
 	obj := map[string]string{}
 	b, err := l.Read(configFile)
 	if err != nil {
+		log.Error(err)
```
abiosoft, line 18: This is not meant to be displayed to the user, as it is expected to fail silently during initial run.
As you can see the error is also discarded on line 24 below.

If at all it would be displayed, the log level should be trace.

environment/vm/lima/config.go
```diff
@@ -9,9 +10,13 @@ import (
 const configFile = "/etc/colima/colima.json"
 
 func (l limaVM) getConf() map[string]string {
+	log := l.Logger(context.Background())
+
 	obj := map[string]string{}
 	b, err := l.Read(configFile)
 	if err != nil {
+		log.Tracef("failed to read config file %v", err)
```
abiosoft, line 18:
```suggestion
		log.Trace(fmt.Errorf("error reading config file: %w", err))
```

abiosoft (approved):

abiosoft:
Thanks :)

PR #1254, opened 2025-01-22
Update README.md

chandler-barlow:
This is the correct nix install command

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ brew install colima
 sudo port install colima
 
 # Nix
-nix-env -iA nixpkgs.colima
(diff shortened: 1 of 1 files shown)


66Ton99:
I do not think it is correct https://search.nixos.org/packages?channel=24.11&show=colima&from=0&size=50&sort=relevance&type=packages&query=colima

abiosoft:
Looking at the official NixOS installation instructions, the current documentation is indeed accurate.

I would be therefore be closing this PR to stay aligned with the official instructions.

Thanks :)

PR #1255, opened 2025-01-24
core: config initial value for kubernetes version if empty in yaml

jason19970210:
According to facing the error while using the `containerd` as runtime and enable `kubernetes` features as the following:

```
FATA[0018] error provisioning kubernetes: error at 'downloading and installing': error downloading 'https://github.com/k3s-io/k3s/releases/download//k3s': error validating SHA sum for 'k3s': exit status 1 
```
It turns out missing `version` string in baseURL which goes with `environment/container/kubernetes/k3s.go`

Thus make a if-else statement for setup `DefaultVersion` which declares in `environment/container/kubernetes/kubernetes.go`

### DIFF
--- a/environment/container/kubernetes/kubernetes.go
+++ b/environment/container/kubernetes/kubernetes.go
@@ -104,6 +104,12 @@ func (c *kubernetesRuntime) Provision(ctx context.Context) error {
 		conf = c.config()
 	}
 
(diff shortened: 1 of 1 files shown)


abiosoft (approved):
Thanks :)

PR #1267, opened 2025-02-05
chore: expend the path information about current profile being used

jason19970210:
As subject mentioned, according to mistaken config yaml which modified manually, it might be showing the full path about the current profile being used with command `$ colima start` as well.


abiosoft:
You have reverted the change, right?

jason19970210:
I think it is proper to renew the PR. Close this one first.

PR #1270, opened 2025-02-10
chore: expend the path information about current profile being used

jason19970210:
Revision of #1267.

As subject mentioned, according to mistaken config yaml which modified manually, it might be showing the full path about the current profile being used with command `$ colima start` as well.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -99,7 +99,8 @@ func (c colimaApp) startWithRuntime(conf config.Config) ([]environment.Container
 func (c colimaApp) Start(conf config.Config) error {
 	ctx := context.WithValue(context.Background(), config.CtxKey(), conf)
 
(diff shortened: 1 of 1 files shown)


abiosoft:
I am actually not a fan of this approach as it is not a relevant information for most users.

I would leave it as is, and log the config file as a trace i.e. with `log.Trace`. If for troubleshooting reasons there is a need to verify the config file being loaded, trace log can be enabled with the `--very-verbose` flag.

abiosoft:
@jason19970210 do you still want to update this?

jason19970210:
@abiosoft I have made change to `log.Tracef` which comes up with "very-verbose".

abiosoft:
@jason19970210 I added one more code comment suggestion.

jason19970210:
@abiosoft it looks good about the suggestions.

abiosoft (approved):

PR #1145, opened 2024-09-27
docker: replace 127.0.0.1 with host gateway ip in docker daemon.json proxy settings

rteeling-evernorth:
- Automatically replace 127.0.0.1 with host gateway ip in docker daemon.json proxy settings
- Refactored daemon.json to move fetching of gateway ip to a function for DRY

Fixes #1144

Related to #956, #1040

Issue #956, opened 2024-01-12
colima x86_64 daemon.json file has incorrect proxy


dragonfriend0013:
### Description

I use zscaler on my machine (M2 Mac), so my local proxy is set to 127.0.0.1:9000.  during the colima start, it set the proxy variable to a 192.* address, but the daemon.json file still shows as the original 127 address.  this should be updates with the updated address.  i can manually update this through ssh, but it would be better if this is done when colima starts.

docker build does not have this issue, only docker compose and docker-compose

### Version

colima version 0.6.7
git commit: ba1be00e9aec47f2c1ffdacfb7e428e465f0b58a

runtime: docker
arch: x86_64
client: v24.0.7
server: v24.0.7
limactl version 0.19.1
qemu-img version 8.2.0
Copyright (c) 2003-2023 Fabrice Bellard and the QEMU Project developers

### Operating System

- [ ] macOS Intel <= 13 (Ventura)
- [ ] macOS Intel >= 14 (Sonoma)
- [ ] Apple Silicon <= 13 (Ventura)
- [X] Apple Silicon >= 14 (Sonoma)
- [ ] Linux

### Output of `colima status`

INFO[0000] colima is running using QEMU
INFO[0000] arch: x86_64
INFO[0000] runtime: docker
INFO[0000] mountType: sshfs
INFO[0000] socket: unix:///Users/jamurphy/.colima/default/docker.sock

### Reproduction Steps

1. colima delete
2. colima start --arch x86_64
3. docker compose up --detach (have an external image that must be pulled through proxy)

### Expected behaviour

docker compose works

### Additional context
env variable on Mac:
http_proxy=http://127.0.0.1:9000
https_proxy=http://127.0.0.1:9000

env variable on colima
https_proxy=http://192.168.5.2:9000
http_proxy=http://192.168.5.2:9000

daemon.json on colima
cat daemon.json
{
  "exec-opts": [
    "native.cgroupdriver=cgroupfs"
  ],
  "features": {
    "buildkit": true
  },
  "proxies": {
    "http-proxy": "http://127.0.0.1:9000",
    "https-proxy": "http://127.0.0.1:9000"
  }
}
_No response_

abiosoft:
It is an oversight. `127.0.0.1` proxy host address should be auto-translated to `192.168.5.2`.

fralken:
Is it necessary to set the proxies in `daemon.json`? In colima 0.5.6 they are not set but `docker info` shows the values from the env variables.

[...]

PR #1298, opened 2025-04-08
docker: replace also localhost with host gateway ip in docker daemon.json

fralken:
This improves PR #1145 by replacing any loopback address, not only _127.0.0.1_ but also _localhost_ with the host gateway.
In my case my proxy is configured as localhost, and it listens both on 127.0.0.1 (IPv4) and ::1 (IPv6).

The implementation logic is the same as in [lima](https://github.com/lima-vm/lima/blob/0625d0b084450e874869dcbc9f63d4312797c3fe/pkg/cidata/cidata.go#L73).

This addresses issue #956.

In my opinion in _no-proxy_ env variable references to localhost should not be replaced, because in this case it refers to the loopback of the container.

### DIFF
--- a/environment/container/docker/daemon.go
+++ b/environment/container/docker/daemon.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
-	"strings"
+	"net/url"
 )
 
 const daemonFile = "/etc/docker/daemon.json"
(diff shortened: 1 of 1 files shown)


environment/container/docker/daemon.go
```diff
@@ -29,6 +29,28 @@ func getHostGatewayIp(d dockerRuntime, conf map[string]any) (string, error) {
...
+		return hostProxy
+	}
+	for _, ip := range ips {
+		if ip.IsLoopback() {
+			newHost := hostGateway
+			if u.Port() != "" {
+				newHost = net.JoinHostPort(newHost, u.Port())
+			}
+			u.Host = newHost
+			hostProxy = u.String()
```
abiosoft, line 48: Should there not be a `break` after this line?

environment/container/docker/daemon.go
```diff
@@ -29,6 +29,28 @@ func getHostGatewayIp(d dockerRuntime, conf map[string]any) (string, error) {
 	return ip, nil
 }
 
+func resolveHostProxy(hostProxy string, hostGateway string) (string) {
```
abiosoft, line 32:
```suggestion
func resolveHostProxy(hostProxy string, hostGateway string) string {
```

fralken: Corrected.

PR #1326, opened 2025-05-22
github workflows: pin dependencies

thypon:
in order to avoid supply chain issues, pin all the dependencies.

Context https://x.com/ramimacisabird/status/1903068773411631426


abiosoft (approved):
Thanks :)

PR #1374, opened 2025-08-07
k3s: retry ip address retrieval during setup

### LINKED ISSUE #1373: k3s installation potentially starts before IP assignment
### Description

When running colima with `--kubernetes`, the VM IP is passed to k3s in the `--advertise-address` parameter. However, the IP may not yet have actually been assigned to the network interface, which means that `--advertise-address` is called without any IP which causes the service to fail with:

```
Aug 07 09:08:56 colima k3s[5836]: Error: invalid argument "--flannel-iface" for "--advertise-address" flag: failed to parse IP: "--flannel-iface"
```

### Version

colima version 0.8.3
git commit: 6a6da99761f300110dfde5ecc789afa35941c3ac
limactl version 1.2.1

### Operating System

- [ ] macOS Intel <= 13 (Ventura)
- [ ] macOS Intel >= 14 (Sonoma)
- [ ] Apple Silicon <= 13 (Ventura)
- [x] Apple Silicon >= 14 (Sonoma)
- [ ] Linux

### Output of `colima status`

INFO[0000] colima [profile=kubernetes] is running using macOS Virtualization.Framework
INFO[0000] arch: aarch64
INFO[0000] runtime: docker
INFO[0000] mountType: virtiofs
INFO[0000] address: 192.168.64.6
INFO[0000] docker socket: unix:///$HOME/.colima/kubernetes/docker.sock
INFO[0000] containerd socket: unix:///$HOME/.colima/kubernetes/containerd.sock

### Reproduction Steps

1. Create a new VM with k3s enabled: `colima start -k -p <profile>`
2. Be unlucky
3. See that the VM comes up but you cannot connect to it with `kubectl`:

```
$ kubectl get pods
E0806 16:47:11.670843    1573 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://192.168.64.6:59300/api?timeout=32s\": dial tcp 192.168.64.6:59300: connect: no route to host"
```

### Expected behaviour

The k3s installation should wait for a certain period of time for an IP to be assigned to the interface, rather than assume that it will always have been assigned already.

### Additional context

_No response_

Nevon:
When networking is enabled but the interface does not have any IP assigned yet, k3s installation can fail. The `IPAddress` function says that it is guaranteed to return a value, which is technically true, but that value can be an empty string if the network interface exists but doesn't have any IP assigned.

This changes the k3s installation to wait for up to 20 seconds for an address to be assigned before failing.

Fixes #1373

### DIFF
--- a/environment/container/kubernetes/k3s.go
+++ b/environment/container/kubernetes/k3s.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/abiosoft/colima/cli"
 	"github.com/abiosoft/colima/config"
@@ -156,18 +157,24 @@ func installK3sCluster(
 		"--write-kubeconfig-mode", "644",
 	}, k3sArgs...)
 
-	// replace ip address if networking is enabled
-	ipAddress := limautil.IPAddress(config.CurrentProfile().ID)
-	if ipAddress == "127.0.0.1" {
-		args = append(args, "--flannel-iface", "eth0")
-	} else {
-		if !hasK3sArg(k3sArgs, "--advertise-address") {
-			args = append(args, "--advertise-address", ipAddress)
+	a.Retry("waiting for VM IP address", time.Second*2, 10, func(retryCount int) error {
+		ipAddress := limautil.IPAddress(config.CurrentProfile().ID)
+		if ipAddress == "" {
+			return fmt.Errorf("no IP address assigned to network interface")
 		}
-		if !hasK3sArg(k3sArgs, "--flannel-iface") {
-			args = append(args, "--flannel-iface", limautil.NetInterface)
+
+		if ipAddress == "127.0.0.1" {
+			args = append(args, "--flannel-iface", "eth0")
+		} else {
+			if !hasK3sArg(k3sArgs, "--advertise-address") {
+				args = append(args, "--advertise-address", ipAddress)
+			}
+			if !hasK3sArg(k3sArgs, "--flannel-iface") {
+				args = append(args, "--flannel-iface", limautil.NetInterface)
+			}
 		}
-	}
+		return nil
+	})
 
 	switch containerRuntime {
 	case docker.Name:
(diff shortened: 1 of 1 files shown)


environment/container/kubernetes/k3s.go
```diff
@@ -156,18 +157,24 @@ func installK3sCluster(
...
 
-	// replace ip address if networking is enabled
-	ipAddress := limautil.IPAddress(config.CurrentProfile().ID)
-	if ipAddress == "127.0.0.1" {
-		args = append(args, "--flannel-iface", "eth0")
-	} else {
-		if !hasK3sArg(k3sArgs, "--advertise-address") {
-			args = append(args, "--advertise-address", ipAddress)
+	a.Retry("waiting for VM IP address", time.Second*2, 10, func(retryCount int) error {
+		ipAddress := limautil.IPAddress(config.CurrentProfile().ID)
```
Nevon, line 161: Something I'm not sure about is whether the actual network interface is guaranteed to exist at this point. Otherwise this will just end up falling back to 127.0.0.1. That behavior is however unchanged by this.

Nevon:
Looks like one of the runners had an issue for one of the checks, but I don't have a way to retrigger it.

environment/container/kubernetes/k3s.go
```diff
@@ -156,18 +157,24 @@ func installK3sCluster(
...
 	}, k3sArgs...)
 
-	// replace ip address if networking is enabled
-	ipAddress := limautil.IPAddress(config.CurrentProfile().ID)
-	if ipAddress == "127.0.0.1" {
-		args = append(args, "--flannel-iface", "eth0")
-	} else {
-		if !hasK3sArg(k3sArgs, "--advertise-address") {
-			args = append(args, "--advertise-address", ipAddress)
+	a.Retry("waiting for VM IP address", time.Second*2, 10, func(retryCount int) error {
```
abiosoft, line 160: why not simply retry 3 times at 5 seconds interval?

```suggestion
	a.Retry("waiting for VM IP address", time.Second*5, 4, func(retryCount int) error {
```

Nevon: Yeah that's fine by me. I just arbitrarily picked 2 seconds because I figured most of the time the IP should be assigned very soon after this check, but that's totally just a guess.

abiosoft (approved):

abiosoft:
Thanks again :)

PR #1391, opened 2025-09-06
config: consider MountPoint in checkOverlappingMounts

sakkyoi:
# Description
Fix overlapping mount check to consider MountPoint

# Problem
The original `checkOverlappingMounts` function only checks for the `Location` field, ignoring `MountPoint`.
This causes valid mount configurations to fail when the `MountPoint` is explicitly set. For example:

```yaml
mounts:
  - location: ~/something/to/mount
    mountPoint: /opt/mount/point
    writable: true
  - location: ~/
    writable: true
```

Currently, this triggers an "overlapping mount" error even though the actual mount points do not overlap.

# Solution
Update `checkOverlappingMounts` to consider `MountPoint` in addition to `Location` when detecting overlapping mounts.
Previously, only `Location` was checked, which could incorrectly fail or allow invalid mounts when a custom `MountPoint` was used.

### DIFF
--- a/environment/vm/lima/yaml.go
+++ b/environment/vm/lima/yaml.go
@@ -389,28 +389,25 @@ func newConf(ctx context.Context, conf config.Config) (l limaconfig.Config, err 
 
 type Arch = environment.Arch
 
+func selectPath(m config.Mount) (string, error) {
+	if m.MountPoint != "" {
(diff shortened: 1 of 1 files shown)


abiosoft (dismissed):

abiosoft (approved):

abiosoft:
Thanks :)

Issue #1434, opened 2025-10-19
Add possibility of specifying the gateway address


Guilospanck:
### Description

When connected to a VPN that has the same Gateway (192.168.5.2) as colima/lima, I couldn't reach out to services in the other side from inside the container. Guess is the conflict of addresses.

- VPN has the same gateway address as colima (192.168.5.2);
- Tries to pull docker image from private registry only accessible via the VPN, which needs to pass through the gateway;
- Not possible.

P.S.: I have made some changes [here](https://github.com/Guilospanck/colima/pull/1) and passing a different GW fixed my problem when using this version, still not sure if this is even needed. If it is, glad to contribute.

abiosoft:
Hi, thanks for taking a dig at working around the issue. 

Your solution is actually accurate (or at least close enough). I am a bit occupied at the moment and would have another look soon :)

abiosoft:
@Guilospanck can you create a pull request?

Guilospanck:
Done @abiosoft

PR #1447, opened 2025-11-11
net: add ability to customize gateway address

Guilospanck:
This allows the possibility of specifying the gateway address when starting colima.

### Issue ([#1434](https://github.com/abiosoft/colima/issues/1434))

When connected to a VPN that has the same Gateway (192.168.5.2) as colima/lima, I couldn't reach out to services in the other side from inside the container. Guess is the conflict of addresses.

- VPN has the same gateway address as colima (192.168.5.2);
- Tries to pull docker image from private registry only accessible via the VPN, which needs to pass through the gateway;
- Not possible.

### Proposed solution

Add a new flag/config to allow specifying the gateway address you want colima to have. If none is passed, it defaults to 192.168.5.2. This prevents the conflicts as above. In my testings it worked for my situation and for other parts that don't depend on the VPN as well.

The gateway must be a valid IPv4 and terminated with 2.

### DIFF
--- a/config/configmanager/configmanager.go
+++ b/config/configmanager/configmanager.go
@@ -124,12 +124,8 @@ func validateGatewayAddress(gateway net.IP) error {
 		return fmt.Errorf("gateway %q is not IPv4", gateway)
 	}
 
-	parts := strings.Split(gateway.String(), ".")
-	if len(parts) != 4 {
-		return fmt.Errorf("gateway %q does not have 4 octets", gateway)
-	}
-
-	if parts[3] != "2" {
+	// Check last octet
+	if ip4[3] != 2 {
 		return fmt.Errorf("the last octet of gateway %q is not 2", gateway)
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/start.go
```diff
@@ -210,6 +211,9 @@ func init() {
 		}
 	}
 
+	// Gateway Address
+	startCmd.Flags().StringVar(&startCmdArgs.Network.GatewayAddress, "gateway-address", "192.168.5.2", "gateway address")
```
abiosoft: You can leverage ip address flag type to avoid manual validation later.

```suggestion
	startCmd.Flags().IPVar(&startCmdArgs.Network.GatewayAddress, "gateway-address", net.ParseIP("192.168.5.2"), "gateway address")
```

config/config.go
```diff
@@ -88,6 +88,7 @@ type Network struct {
 	Mode            string            `yaml:"mode"` // shared, bridged
 	BridgeInterface string            `yaml:"interface"`
 	PreferredRoute  bool              `yaml:"preferredRoute"`
+	GatewayAddress  string            `yaml:"gatewayAddress"`
```
abiosoft:
```suggestion
	GatewayAddress  net.IP            `yaml:"gatewayAddress"`
```

abiosoft:
I've been away for a while but I am ready to get this merged :)
Kindly attend to the comments.

Guilospanck:
@abiosoft Improved it based on your comments

abiosoft:
Thanks :)

PR #1487, opened 2026-02-03
incus: use RFC 5737 IPv4 network for the default network

### LINKED ISSUE #1486: Incus network configuration uses publicly routable IPv4 network as a private network
### Description

The Incus config colima uses [here](https://github.com/abiosoft/colima/blob/ccd5ddd03ee07373160464840e747e5e0e587da1/environment/container/incus/config.yaml#L3) defines the Incus private network as `192.100.0.0/24`. This is not within the private IP address space specified by IANA in [RFC 1918](https://www.rfc-editor.org/rfc/rfc1918), and is currently allocated to an organization called NetDocuments. Was this meant to be `192.168.100.1` instead?

### Version

```colima version 0.9.1
limactl version 2.0.3
qemu-img version 10.2.0
Copyright (c) 2003-2025 Fabrice Bellard and the QEMU Project developers
```

### Operating System

- [ ] macOS Intel <= 13 (Ventura)
- [ ] macOS Intel >= 14 (Sonoma)
- [ ] Apple Silicon <= 13 (Ventura)
- [x] Apple Silicon >= 14 (Sonoma)
- [ ] Linux

### Output of `colima status`

N/A

### Reproduction Steps

N/A

### Expected behaviour

The Incus private network should fall in the IANA-allocated private network space.

### Additional context

_No response_

jessa0:
[RFC 5737](https://datatracker.ietf.org/doc/html/rfc5737) establishes a few IP subnets reserved for documentation-use. Since in the comments for 3317db1 it was expressed that users should simply run `incus network edit incusbr0`, it seems appropriate that a documentation-use range be used as the default here, since the defaults then kind-of represent a sample config which should be changed later.

It's also not the end of the world if two applications clash in the RFC5737 range, since if that happens, the user should simply change one of the applications to a subnet of their choice. The RFC5737 probably will never be a system-default subnet that could cause bad behaviour if clashed with.

Fixes #1486.

### DIFF
--- a/environment/container/incus/config.yaml
+++ b/environment/container/incus/config.yaml
@@ -1,6 +1,6 @@
 networks:
   - config:
-      ipv4.address: 192.100.0.1/24
+      ipv4.address: 192.168.100.1/24
       ipv4.nat: "true"
       ipv6.address: auto
     description: ""


jessa0:
Sorry, just a moment, need to change `192.0.2.0/24` to `192.0.2.1/24`.

abiosoft:
Kindly sign your commits.
You can see the instructions [here](https://github.com/abiosoft/colima/blob/main/docs/CONTRIBUTE.md#3-commit-your-changes).

jessa0:
> Kindly sign your commits.

fixed in 69b3f7c

abiosoft:
Thanks :)

PR #1509, opened 2026-02-12
added install instructions for krunkit to avoid AI mode colima start

alanpmullane:
added install instructions for krunkit to avoid AI mode colima start

Alan Mullane <alan.p.mullane@gmail.com>


abiosoft:
Try commiting with `git commit -s`.

PR #1523, opened 2026-02-27
containerd: support user-provided containerd and buildkit config

### LINKED ISSUE #212: Support patching base containerd config
I am working with colima to develop local development workflows in golang. One of the challenges here is that the VM that colima spins up from https://github.com/abiosoft/alpine-lima has the base containerd config file present on the host at startup time. This means the generated `lima.yaml` configfile has the containerd settings turned off and does not do the default overrides present in https://github.com/lima-vm/lima/blob/master/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh.

I'd love to have a workflow that looks like the kind cluster patches that allows me to configure the containerd plugin settings: https://kind.sigs.k8s.io/docs/user/local-registry/#create-a-cluster-and-registry. I can definitely help contribute here if pointed in the right direction, but wanted to get some input from the maintainers before proceeding.

I am also using an M1 Mac, and I'm developing workflows on top of colima. Would love to help testing in certain areas of need!

citadelgrad:
## Summary

- Adds support for user-customizable containerd `config.toml` and buildkit `buildkitd.toml` config files, matching the existing Docker `daemon.json` pattern
- Uses standard XDG config locations (`~/.config/containerd/` and `~/.config/buildkit/`) as the central config, shared across all Colima profiles
- Supports per-profile overrides at `~/.colima/<profile>/containerd/` for users who need different configs per instance
- On first run, default configs are written to the central location for user discovery and editing
- Adds documentation to FAQ.md explaining the config customization workflow

Closes #212

## Details

The containerd runtime previously always wrote the embedded default configs to the VM with no way for users to customize them. This was a gap compared to the Docker runtime, which supports `daemon.json` customization.

Config files are resolved in the following order:
1. **Per-profile override** at `~/.colima/<profile>/containerd/<file>` — for users who need different configs per instance
2. **Central config** at `~/.config/containerd/config.toml` or `~/.config/buildkit/buildkitd.toml` — shared across all profiles, follows the standard rootless containerd/buildkit conventions
3. **Embedded default** — written to the central location on first run for discoverability

This approach:
- Provides a single central config location (addressing the per-instance duplication concern)
- Follows established rootless containerd/buildkit XDG conventions
- Respects `$XDG_CONFIG_HOME` if set
- Still allows per-profile overrides for power users

## LLM Disclosure

This PR was authored with assistance from Claude (Anthropic). The Go code changes (`provisionConfig()` and `userConfigDir()` in `containerd.go`), unit tests, and FAQ documentation were drafted by Claude based on analysis of the existing Docker `daemon.json` pattern in `daemon.go`, issue #212 discussion, maintainer feedback, and the codebase architecture. All code was reviewed for correctness and verified with `go build ./...` and `go test ./...`.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes — all existing tests plus 4 new unit tests
- [x] Unit tests verify three-tier config resolution:
  - Profile override takes priority over central and default
  - Central config is used as fallback when no profile override exists
  - Embedded default is written to central location on first run
  - `userConfigDir()` respects `$XDG_CONFIG_HOME`
- [x] `colima start --runtime containerd` generates default configs at `~/.config/containerd/config.toml` and `~/.config/buildkit/buildkitd.toml`
- [x] Editing the central config and restarting picks up the changes in the VM
- [x] Per-profile override at `~/.colima/<profile>/containerd/config.toml` takes priority over the central config
- [x] Docker runtime is unaffected

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Set up Go
-        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
         with:
           go-version: "1.25.1"
 
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Set up Go
-        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
         with:
           go-version: "1.25.1"
 
@@ -57,7 +57,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Set up Go
-        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
         with:
           go-version: "1.25.1"
 
@@ -72,7 +72,7 @@ jobs:
           OS=Linux ARCH=aarch64 make
 
       - name: upload artifacts
-        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: artifacts-linux
           path: _output/binaries/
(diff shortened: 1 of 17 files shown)


abiosoft:
Thanks for your efforts.

The main issue I have with this is the need to have config file(s) per instance. Unlike docker that has a central location for the config file.

Any idea if there is a standard config file location within $HOME for containerd that we can leverage, similar to docker?

citadelgrad:
Thanks for the feedback! I've reworked the approach based on your concerns.

**What changed:** Instead of per-instance config files only, the PR now uses a three-tier resolution with a central config location following containerd/buildkit's own rootless conventions:

1. **Per-profile override** at `~/.colima/<profile>/containerd/config.toml` (highest priority)
2. **Central config** at `~/.config/containerd/config.toml` (shared across all instances)
3. **Embedded default** (written to the central location on first run for discoverability)

The central path follows the XDG base directory spec (`$XDG_CONFIG_HOME/containerd/config.toml`, defaulting to `~/.config/containerd/config.toml`), which is the same convention containerd and buildkit use for their rootless configurations. Same pattern for buildkit at `~/.config/buildkit/buildkitd.toml`.

This means most users just edit one file and it applies to all Colima instances, while per-profile overrides are still available when needed.

I also added unit tests covering all three resolution paths and updated the FAQ docs.

abiosoft:
You can actually get rid of the tests, it's unnecessarily verbose with little benefits.

Also, kindly sign your commits.

citadelgrad:
Removed my tests and signed my last commit. Do you need me to resign the other commits?

abiosoft:
> Removed my tests and signed my last commit. Do you need me to resign the other commits?

The DCO check is failing. You can either sign all commits, or rebase and make it a single signed commit.

PR #1522, opened 2026-02-27
docs: add VM customization and update guide to FAQ

citadelgrad:
## Summary

- Expands the "Are Lima overrides supported?" FAQ section with concrete examples for `override.yaml` location and provision scripts (both via Lima overrides and `colima.yaml`)
- Adds a new "How can the VM and its tools be updated?" section covering `brew upgrade`, `colima update` (v0.7.6+), `colima ssh`, and environment variables (`COLIMA_HOME`, `COLIMA_PROFILE`, `DOCKER_HOST`, `--env`)
- Updates the table of contents

Closes #105

## LLM Disclosure

This PR was authored with assistance from Claude (Anthropic). The FAQ content was drafted by Claude based on existing Colima documentation, issue discussions, and codebase analysis. All content was reviewed for accuracy before submission.

## Test plan

- [x] Verified all internal anchor links resolve correctly on GitHub
- [x] Verified markdown renders correctly on GitHub

PR #1537, opened 2026-03-08
core: make cache directory configurable

reillylm:
I routinely run out of space on my Macbook's 256GB drive. I have `COLIMA_HOME` set on an external drive to hold the Lima disks, but it would be nice to have the cache directory there too.

I wasn't sure if we wanted to require the directory already exist (as we do with `$COLIMA_HOME`) or create the directory automatically (as we do with `$LIMA_HOME`), so I picked the latter (otherwise, it'll fall back silently to the default, which feels wrong if we're explicitly setting the env var).

Noted the new env var down in the FAQ file. Also moved where we talk about the env vars from #1522, since I thought it was a little confusing how it was written. Specifically, `COLIMA_HOME` / `COLIMA_PROFILE` / `DOCKER_CONFIG` are env vars for the host to configure the Colima CLI (or Docker CLI), whereas `--env` is used to inject env vars into the guest VM, so I tried to make that more clear.

### DIFF
--- a/config/files.go
+++ b/config/files.go
@@ -100,8 +100,11 @@ var (
 
 	cacheDir = requiredDir{
 		dir: func() (string, error) {
-			dir := os.Getenv("XDG_CACHE_HOME")
-			if dir != "" {
+			if dir := os.Getenv("COLIMA_CACHE_HOME"); dir != "" {
+				return dir, nil
+			}
+
+			if dir := os.Getenv("XDG_CACHE_HOME"); dir != "" {
 				return filepath.Join(dir, "colima"), nil
 			}
 			// else
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -11,6 +11,8 @@
     - [Editing the config](#editing-the-config)
     - [Setting the default config](#setting-the-default-config)
     - [Specifying the config editor](#specifying-the-config-editor)
+  - [How do I change where Colima files are stored?](#how-do-i-change-where-colima-files-are-stored)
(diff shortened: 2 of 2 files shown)


reillylm:
@abiosoft side note: I noticed that the [FAQ page in the docs](https://colima.run/docs/faq) and the `FAQ.md` file in the repo don't exactly match. Also that the rest of the docs pages from the website are not in the repo. For example, for env vars, I noticed that the `env` YAML config property is already covered in the [Configuration page on the docs](https://colima.run/docs/configuration/#environment-variables), so I'm not sure if you wanted to just remove the FAQ entry altogether?

abiosoft:
The cache directory is actually a special case as it is always mounted in the VM, downloaded artifacts needed in the VM are retrieved via the mounted cache directory. Therefore, locating it on external disk would remove the guarantee of it always being mounted.

Further updates are needed in that regard before this PR can be merged.

As for the FAQ, yeah, the website is more detailed. I plan to eventually get rid of the FAQ.md file and direct users to the website instead.

abiosoft:
FYI, you can clear the cache (to save disk space) with `colima prune` command.

abiosoft:
@reillylm you can do a rebase and we should be good to go 👍🏽

reillylm:
Hmm interesting, I thought the cache directory was only used when creating a VM the first time to download the disk image?

abiosoft (approved):

abiosoft:
> Hmm interesting, I thought the cache directory was only used when creating a VM the first time to download the disk image?

It was used for more than that.

abiosoft:
Thanks :)

PR #1589, opened 2026-06-14
cli: add --registry-mirror flag to start command

### LINKED ISSUE #1445: Add registry mirror flag to start command
### Description

I don't edit the template file since my colima setup is all cli usage based and pretty simple, I just set cpu and memory on start.

A start flag for adding a registry mirror entry would be great. I'm sure many folks would add gcr mirror too!

```sh
colima start --registry-mirror "https://mirror.gcr.io" --registry-mirror etc...
```

Thank you.

ekalinin:
Add a repeatable --registry-mirror flag as a CLI convenience for configuring docker registry mirrors without editing the config file. The values are injected into the docker daemon.json registry-mirrors key, which was previously only settable via the `docker:` config block.

Closes #1445

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -475,6 +475,9 @@ func prepareConfig(cmd *cobra.Command) {
 	startCmdArgs.ActivateRuntime = &startCmdArgs.Flags.ActivateRuntime
 	startCmdArgs.Binfmt = &startCmdArgs.Flags.Binfmt
 	startCmdArgs.ForceDiskImage = &startCmdArgs.Flags.ForceDiskImage
+	if cmd.Flag("registry-mirror").Changed {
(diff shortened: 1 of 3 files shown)


cmd/start.go
```diff
@@ -485,15 +499,21 @@ func prepareConfig(cmd *cobra.Command) {
 
 		if !templateUsed {
 			// use default config if there is no template or template is disabled
+			if cmd.Flag("registry-mirror").Changed {
+				startCmdArgs.Docker = withRegistryMirrors(startCmdArgs.Docker, startCmdArgs.Flags.RegistryMirrors)
+			}
```
abiosoft: This should be positioned up top around the `convert cli to config file format` section.

ekalinin: Good point. Thanks, fixed.

abiosoft (approved):

abiosoft:
Thanks :)

PR #1591, opened 2026-06-15
core: add --disk-image-mirror flag to download images from a mirror

### LINKED ISSUE #1438: Enable downloading qcow2 images from a mirror
### Description

The URLs of the prepackaged VMs are stored in [./embedded/images/images.txt](https://github.com/abiosoft/colima/blob/main/embedded/images/images.txt) along with the hash. That's great for security, but not great for folks who would prefer to retrieve these 220-520 MiB files from a mirror.

For my particular use case, being able to replace `https://github.com` with `https://artifactory.mycompany.com/artifactory/github` would suffice.

Poking around a bit, it seems like primarily these functions would have to change, perhaps to account for an environment variable allowing prefixing a hostname or filepath. 

https://github.com/abiosoft/colima/blob/c5e126abe6964e9839d3c27cc50b313677c5d150/environment/vm/lima/limautil/image.go#L101-L139

ekalinin:
Allow replacing the https://github.com prefix of the disk image URL with a mirror (e.g. an Artifactory proxying GitHub releases). The sha512 checksum is still validated after download, so a mirror cannot serve tampered images.

Closes #1438

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -194,6 +194,7 @@ func init() {
 	startCmd.Flags().BoolVarP(&startCmdArgs.Flags.Foreground, "foreground", "f", false, "Keep colima in the foreground")
 	startCmd.Flags().StringVar(&startCmdArgs.Hostname, "hostname", "", "custom hostname for the virtual machine")
 	startCmd.Flags().StringVarP(&startCmdArgs.DiskImage, "disk-image", "i", "", "file path to a custom disk image")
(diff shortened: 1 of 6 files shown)


abiosoft (approved):

abiosoft:
Thanks :)

colindean:
@ekalinin Thank you so very much for doing this work! I _really_ appreciate it.

@abiosoft As always, thank you :pray:

PR #1609, opened 2026-07-24
core: update mount type for krunkit VM to virtiofs when 9p is specified

Kit09K:
### Description

Currently, starting Colima with `--vm-type krunkit` fails immediately if `--mount-type` is not explicitly set, because the configuration defaults to `9p`. The krunkit driver only supports `virtiofs` or `reverse-sshfs`, resulting in the following fatal error:

`rpc error: code = Unknown desc = field mountType must be virtiofs or reverse-sshfs for krunkit driver, got 9p`

### Changes

This PR adds the missing fallback logic in `cmd/start.go` (inside `setFlagDefaults`). It automatically updates the `mountType` to `virtiofs` when `krunkit` is used and `9p` is currently specified. 

This mirrors the existing behavior that is already implemented for the `vz` vmType, allowing users to start a krunkit VM smoothly without manually overriding the mount type.

PR #1610, opened 2026-07-28
core: fix krunkit virtiofs configuration

jasonthird:
## What changed

- Treat krunkit like VZ when generating Lima's default mount type.
- Keep explicit SSHFS aliases mapped to `reverse-sshfs`.
- Add platform-independent regression coverage for QEMU, VZ, and krunkit mount resolution.

## Why

The CLI now converts krunkit's unsupported `9p` mount type to `virtiofs`, but the Lima configuration generator still treated every non-VZ backend as QEMU and emitted `9p`. As a result, starting krunkit failed Lima validation even when `virtiofs` was requested.

This change keeps the generated Lima configuration consistent with the CLI behavior added in #1609.

## Impact

`colima start --vm-type krunkit --mount-type virtiofs` now generates a valid `virtiofs` Lima configuration. Existing QEMU, VZ, and SSHFS behavior is preserved.

## Validation

- `go test ./...`
- `gofmt`
- `git diff --check`

## LLM usage disclosure

OpenAI Codex using GPT-5.6-sol was used to investigate the root cause, implement the focused mount-type resolver change, and generate the regression tests. The resulting diff was inspected, formatted, and validated with the full Go test suite.

### DIFF
--- a/environment/vm/lima/yaml.go
+++ b/environment/vm/lima/yaml.go
@@ -339,16 +339,7 @@ func newConf(ctx context.Context, conf config.Config) (l limaconfig.Config, err 
 		}
 	}
 
-	switch strings.ToLower(conf.MountType) {
-	case "ssh", "sshfs", "reversessh", "reverse-ssh", "reversesshfs", limaconfig.REVSSHFS:
-		l.MountType = limaconfig.REVSSHFS
-	default:
-		if l.VMType == limaconfig.VZ {
-			l.MountType = limaconfig.VIRTIOFS
-		} else { // qemu
-			l.MountType = limaconfig.NINEP
-		}
-	}
+	l.MountType = resolveMountType(l.VMType, conf.MountType)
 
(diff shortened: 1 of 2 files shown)


abiosoft:
Thanks :)

Kindly sign the DCO, you can check [here](https://github.com/abiosoft/colima/blob/main/docs/CONTRIBUTE.md#3-commit-your-changes) for instructions.

oanhnn:
@jasonthird 
Please sign DCO by follow https://github.com/abiosoft/colima/pull/1610/checks?check_run_id=91642353654

It would be great if it were included in the next release.

jasonthird:
> @jasonthird Please sign DCO by follow https://github.com/abiosoft/colima/pull/1610/checks?check_run_id=91642353654
> 
> It would be great if it were included in the next release.

Thanks! I’ve signed off the commits and the DCO check is now passing.

PR #1593, opened 2026-06-17
config: reject mount paths containing spaces

ekalinin:
Mount paths with spaces (e.g. "/Volumes/External HD") fail silently: Colima passes the path to Lima correctly, but the unescaped space breaks the generated /etc/fstab entry and the `nofail` option hides the failure, so the volume is never mounted and no error is shown.

The real fix belongs upstream in Lima. Until then, validate mounts in ValidateConfig and fail fast with a clear error instead of starting with a silently-broken mount, and document the limitation in the FAQ.

Fixes #1471

PR #1615, opened 2026-08-08
config: reject uppercase http(s) diskImage URLs

YuriNachos:
## Summary

`ValidateConfig` rejected a `diskImage` starting with `http://` or `https://`, but the scheme match was case-sensitive. Per [RFC 3986 §3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1) URL schemes are case-insensitive, so `HTTP://`, `Https://`, `hTTpS://`, … slipped past validation and failed much later while booting the VM with an opaque `os.Stat` "file not found" instead of the intended "remote URLs not supported" error.

This makes the http/https scheme match case-insensitive and surfaces it through a focused, directly-testable `validateDiskImage` helper, mirroring the existing `validateMounts` / `validateGatewayAddress` helpers (same shape as #1593).

## Root cause

`strings.HasPrefix(c.DiskImage, "http://")` is byte-exact. `HTTP://example.com/img.qcow2` does not start with the lowercase literal, so `ValidateConfig` returned `nil`. The value was then stored as the Lima image `Location` (`environment/vm/lima/disk.go`), and `os.Stat(conf.DiskImage)` failed during `downloadDiskImage` with a generic bootstrap error that gave no hint the *scheme* was the problem — a worse error than the one the validator existed to produce.

## Changes

| File | Change |
| --- | --- |
| `config/configmanager/configmanager.go` | Extract unexported `validateDiskImage(diskImage string) error`; match the http/https scheme case-insensitively via `strings.ToLower`; call it from `ValidateConfig` in place of the old inline block. Error message is unchanged. |
| `config/configmanager/configmanager_test.go` | Add table-driven `TestValidateDiskImage` covering lowercase, uppercase, and mixed-case http/https, empty, local paths, and the non-http(s) boundary (`ftp://`, and a scheme that merely contains "http"). |
| `docs/FAQ.md` | Add a short "Disk image must be a local file" entry noting the http(s) URL rejection (case-insensitive). |

Validation only — `environment/vm/lima/disk.go` and the rest of the runtime path are untouched. Only `http`/`https` are rejected (other schemes are unchanged behavior); the match is made case-insensitive per RFC 3986 §3.1, nothing more.

## Test plan

- [x] `go test ./config/configmanager/... -run TestValidateDiskImage -v` — all subtests pass (lowercase, uppercase, mixed-case http/https rejected; local paths and `ftp://` allowed)
- [x] `go build ./...`
- [x] `golangci-lint run --timeout 3m` (v2.11.3, `0 issues.`)
- [x] `go test ./...`
- [x] Red-before-green + mutation check: the test fails to compile before the helper exists; temporarily reverting the helper to case-sensitive fails the 5 uppercase/mixed-case subtests, confirming the test guards the fix
- [ ] Verify at runtime on macOS/Apple Silicon (out of scope — this PR is validation-only; the downstream `os.Stat`/download path is unchanged and cannot be exercised headlessly)

No linked issue — self-found robustness fix. Precedent: #1593 (`config: reject mount paths containing spaces`).

### DIFF
--- a/config/configmanager/configmanager.go
+++ b/config/configmanager/configmanager.go
@@ -82,10 +82,8 @@ func ValidateConfig(c config.Config) error {
 		}
 	}
 
-	if c.DiskImage != "" {
-		if strings.HasPrefix(c.DiskImage, "http://") || strings.HasPrefix(c.DiskImage, "https://") {
-			return fmt.Errorf("cannot use diskImage: remote URLs not supported, only local files can be specified")
-		}
+	if err := validateDiskImage(c.DiskImage); err != nil {
+		return err
 	}
 
 	if _, ok := validPortForwarders[c.PortForwarder]; !ok {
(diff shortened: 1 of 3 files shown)


YuriNachos:
A gentle nudge on this small config validation fix — is anything else needed from my side (tests, wording of the error message) to move it forward?

PR #1625, opened 2026-08-20
Add configurable data disk filesystem

### LINKED ISSUE #1624: Allow configuring the FileSystem Format for the Data Disk
### Description

DataDisk FSType is currently hardcoded to Ext4. An underlying filesystem that supports `cp --reflink` (ie xfs or btrfs) would be very useful for some use cases (i.e. running many instances of a postgres db starting from a common seed).

joearasin:
Add a new `diskFS` config option for the VM data disk, defaulting to `ext4` and validating allowed values (`ext4`, `xfs`). 

- Preserve the value in profile defaults and reject unsupported changes after creation
- XFS is limited to profiles whose generated labels fit within the 12-byte Linux limit (which basically means the default profile only, for now, due to label name generation)
- Mount and resize scripts now format and mount the correct filesystem and refuse to reinterpret an existing disk with a different filesystem. 
 
This also updates the default template and adds coverage for disk validation and mount/resize behavior.

See related https://github.com/abiosoft/colima-core/pull/20

Closes #1624

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -67,10 +67,17 @@ Run 'colima template' to set the default configurations or 'colima start --edit'
 		if err != nil {
 			return err
 		}
+		setConfigDefaults(&conf)
+		setFixedConfigs(&conf)
 
 		// validate config
 		if err := configmanager.ValidateConfig(conf); err != nil {
 			return fmt.Errorf("error in config file: %w", err)
+		}
+		if startCmdArgs.Flags.SaveConfig {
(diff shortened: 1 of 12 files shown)


abiosoft:
Thanks for this :)

joearasin:
Also, sorry, did not intend to include 5f823d5 into this PR -- had that for my own use (though you will need something similar when https://github.com/abiosoft/colima-core/pull/20) is merged.

Issue #1490, opened 2026-02-06
Process leak when using custom LaunchAgent with KeepAlive


elfgzp:
## Description

I encountered a process leak issue when running Colima with a custom LaunchAgent configuration. After investigating, I found the cause and wanted to share my findings in case it helps others.

## Environment

- macOS: Darwin 25.2.0 (Apple Silicon)
- Colima version: (installed via Homebrew)
- VM Type: vz

## The Problem

I had created a custom LaunchAgent plist at `~/Library/LaunchAgents/com.github.abiosoft.colima.plist` with the following configuration:

```xml
<key>KeepAlive</key>
<true/>
```

When Colima failed to start (due to an unrelated disk attachment error), the `KeepAlive` setting caused launchd to continuously restart the process. However, each restart attempt spawned new `limactl` processes (especially `limactl usernet`) that were not properly cleaned up when the parent process exited.

Over time, this resulted in **3,800+ orphaned limactl processes** accumulating on my system, which caused:
- Extremely high system load (Load Average: 200-600)
- `kernel_task` consuming significant CPU managing these processes
- Overall system sluggishness

## Error Logs

From `/tmp/colima.err`, the startup was failing repeatedly:
```
time="2026-02-06T11:09:36+08:00" level=fatal msg="error starting vm: error at 'starting': exit status 1"
time="2026-02-06T11:09:47+08:00" level=fatal msg="error starting vm: error at 'starting': exit status 1`
... (repeated ~50 times in 10 minutes)
```

From `~/.colima/_lima/colima/ha.stderr.log`:
```
{"level":"fatal","msg":"failed to run attach disk \"colima\", in use by instance \"colima\"","time":"2026-02-06T11:19:54+08:00"}
```

## Resolution

I noticed that the official Homebrew service configuration uses a different KeepAlive setting:

```xml
<key>KeepAlive</key>
<dict>
    <key>SuccessfulExit</key>
    <true/>
</dict>
```

After updating my LaunchAgent to use `SuccessfulExit` instead of a plain `<true/>`, and cleaning up the orphaned processes, the issue was resolved.

I also found that using `brew services start colima` with the official Homebrew plist works well and avoids this problem entirely.

## Investigation Notes

- The issue appears specific to using `KeepAlive: true` (boolean) versus `KeepAlive.SuccessfulExit: true` (dictionary)
- When startup fails repeatedly, the boolean form causes immediate restart without proper cleanup
[...]

Issue #1265, opened 2025-02-04
`brew services start colima` not working as expected


t18n:
### Description

`brew services start colima` not working as expected. When I run the command and restart the computer, the service does start, but `colima` doesn't.

### Version

❯ colima version && limactl --version && qemu-img --version
colima version 0.8.1
git commit: 96598cc5b64e5e9e1e64891642b91edc8ac49d16
WARN[0000] found ~/.colima, ignoring $XDG_CONFIG_HOME...
WARN[0000] delete ~/.colima to use $XDG_CONFIG_HOME as config directory
WARN[0000] or run `mv ~/.colima "/Users/tt/.config/colima"`

runtime: docker
arch: aarch64

limactl version 1.0.4
qemu-img version 9.2.0
Copyright (c) 2003-2024 Fabrice Bellard and the QEMU Project developers

### Operating System

- [ ] macOS Intel <= 13 (Ventura)
- [ ] macOS Intel >= 14 (Sonoma)
- [ ] Apple Silicon <= 13 (Ventura)
- [x] Apple Silicon >= 14 (Sonoma)
- [ ] Linux

### Output of `colima status`

```
WARN[0000] found ~/.colima, ignoring $XDG_CONFIG_HOME...
WARN[0000] delete ~/.colima to use $XDG_CONFIG_HOME as config directory
WARN[0000] or run `mv ~/.colima "/Users/tt/.config/colima"`
WARN[0000] found ~/.colima, ignoring $XDG_CONFIG_HOME...
WARN[0000] delete ~/.colima to use $XDG_CONFIG_HOME as config directory
WARN[0000] or run `mv ~/.colima "/Users/tt/.config/colima"`
INFO[0000] colima is running using macOS Virtualization.Framework
INFO[0000] arch: aarch64
INFO[0000] runtime: docker
INFO[0000] mountType: sshfs
INFO[0000] socket: unix:///Users/tt/.colima/default/docker.sock
```

### Reproduction Steps

1. Install colima, docker with homebrew
2. Run `colima start` and it works fine
3. Run `brew services start colima` and restart the computer

### Expected behaviour

`brew services start colima` should start Docker service hence start its containers automatically

#### Result of `cat /opt/homebrew/var/log/colima.log`
```
time="2025-02-04T00:28:12+01:00" level=info msg="provisioning ..." context=docker
colima
Successfully created context "colima"
colima
Current context is now "colima"
[...]

Issue #262, opened 2022-04-22
Colima daemon


YentlFrickx:
Currently we always have to manually start colima when starting our macbook, however it would be nice to have it as a daemon so that it can start automatically. Is there any reason why this has not been done yet?

edwardtheharris:
+1

jbreiding:
I am interested in taking this on but curious about what the end result might look like.

The simple path forward is a Colima process per vm, most configurations are likely to only have one and this is how qemu operates. I would think the IPC surface for this is minimal and just operating like detach mode as docker does.
The result is a process to only supervise the qemu process spawned.

Or is there a desire for this to introduce a true daemon that manages all Colima instances?

Thoughts?

jbreiding:
I spent some time this weekend and got to what I would consider a draft PR.

However, this same library is used to launch a few other daemons, and I don't believe this nesting will work.

Qemu exits after complaining about missing the gvproxy domain socket?

Definitely looks a lot like this.
https://github.com/sevlyar/go-daemon/issues/82

fskaeh:
Hello,
Is there any news regarding this issue?
On my system, the drives can take up to a minute to mount at boot, so it would be useful to be able to start colima as a launchd service to wait until the mounting is done.

lloeki:
> The simple path forward is a Colima process per vm

This would be fine, as I don't necessarily want to start all of the ones I have.

Commands like:

- `colima launchagent [profile] [flags]` with flags possibly including `--user` (plist in `~/Library/LaunchAgents`, start on login, stop on logout) or `--system` (plist in `/Library/LaunchAgents`, same but for applies to all system users)
- `colima launchdaemon [profile] [flags]` with flags possibly including `--system` (plist in `/Library/LaunchDaemons`) but not `--user`; kinda redundant though.

Or alternatively a more OS-independent:

```
colima daemon [profile] [--launchagent | --launchdaemon | --systemd] [--user | --system]
```

For macOS although orthogonal, it would need this to be solved: https://github.com/abiosoft/colima/issues/490

lloeki:
I noted that [there's a Homebrew LaunchDaemon](https://github.com/Homebrew/homebrew-core/pull/94189) that appears to have worked at some point. Dunno if it works lately (not a Homebrew user).

I opened an issue at `nix-darwin`: https://github.com/LnL7/nix-darwin/issues/1182

This current issue would still be useful for users that are not using `nix-darwin` though.

[...]

Issue #74, opened 2021-11-30
launchd auto start support missing?


csetera:
The colima stuff is working great as a Docker Desktop replacement with one minor exception for me.  I've been unable to get it to autostart via launchd.  I *think* that it is likely because the "start" command forks the hypervisor and then eventually exits.  My understanding is that launchd expects the command to run indefinitely.  I took a look at the code, but not conversant enough in Go to figure out how to "fix" this.  Is this something that is on the roadmap?

abiosoft:
Yeah, auto-startup has not been added to the project.

csetera:
Thanks.  In the short term, simply making an option that leaves the colima application running in the foreground might be enough to give people a way to wrap it up in launchd.  Anyway... Thanks for this project!

frankievalentine:
> Yeah, auto-startup has not been added to the project.

This would be good to put in the project home README, imo.

Amar1729:
see also #96

lloeki:
Also related:

- https://github.com/abiosoft/colima/issues/262
- https://github.com/abiosoft/colima/issues/490

resker:
FYI — upstream Lima can do this natively now, which should make it straightforward for Colima to adopt.

`limactl autostart enable --condition=boot <instance>` installs a system LaunchDaemon w/ KeepAlive, so it starts at boot rather than at login — no GUI session needed.

The blocker for Colima was that the generated plist carried no LIMA_HOME, so at boot it resolved the default `~/.lima` and never found an instance under `~/.colima/_lima`. Fixed in lima-vm/lima#5489 (merged, slated for Lima v2.3.0):

`LIMA_HOME=~/.colima/_lima limactl autostart enable --condition=boot colima`

It'd seem to be worthwhile to expose this as an option rather than having users hand-roll plists — #1490 shows how that goes wrong, and #262 / #1265 are after the same thing from different angles.

A macOS limitation (by design) that'd represent a caveat to document though: `/Library/LaunchDaemons` is root-owned so the invoking account needs sudo, and` limactl` won't run as root — so `sudo limactl` isn't a way around it.

I'll follow up w/ a PR for consideration.

resker:
Following up on the above — draft PR is #1636 (there are some open questions whether this is approach is desired by the Colima project though)

PR #1636, opened 2026-09-12
cli: add `colima autostart` to start at login or at boot

resker:
Refs #74 (also #262, #1265, #1490).

Lima can register an instance with launchd or systemd to start automatically — either at
login, or on macOS at system boot before anyone logs in. The second is what a headless
machine needs, and is what users have been approximating with hand-rolled plists.

This exposes it as a Colima subcommand:

```sh
colima autostart enable          # at login
colima autostart enable --boot   # at system boot (macOS, prompts for sudo)
colima autostart disable
```

It is a thin delegation to `limactl autostart`. Colima's docker socket is already a
`PortForward` in the generated Lima YAML, so `limactl start <instance>` restores it
without Colima's involvement — no separate plist generation is needed here.

## Draft: depends on an unreleased Lima

A Colima instance lives under `~/.colima/_lima`, not the default `~/.lima`, so this only
works once the generated launchd/systemd unit carries `LIMA_HOME`. That landed in
lima-vm/lima#5489, which is merged but not in any release yet — the current stable is
Lima v2.2.0, and it missed v2.3.0-beta.0. Marking this draft until Lima v2.3.0 ships.

## Open questions

**1. Provision scripts, and the relationship to `brew services`.** This delegates to `limactl autostart`, so the generated unit runs `limactl start <instance> --foreground`. Lima's plist hardcodes its own binary, so there is no way to point the unit at `colima start` instead. The consequence is that the provision scripts Colima runs itself — those with mode `after-boot` or `ready`, per `Provision.IsColimaMode` — do not run on an automatic start. `colima autostart enable` warns when the instance has any, rather than failing silently.

That may well be the wrong trade. `brew services start colima` runs `colima start -f`, the full path, and preserves them. The alternative here is for Colima to generate its own LaunchAgent/LaunchDaemon (and systemd unit) invoking `colima start -f`, which would give parity, drop the Lima version requirement entirely, and make autostart and `brew services` the same path — at the cost of Colima owning plist generation, the sudo write, `launchctl bootstrap`/`bootout`, and the Linux equivalent, all of which Lima maintains today.

I took the smaller option to keep this reviewable, but I suspect you may want the parity. Happy to rework it that way — or to close this if `sudo brew services start colima` covers enough of the need.

**2. Command shape.** `colima autostart enable|disable` mirrors `limactl autostart`. A `colima start --autostart` flag is the alternative.

**3. Profile.** Targeted with the global `--profile` flag rather than a positional arg, since the positional-profile handling in `cmd/root` is a fixed list of command names.

## Known limitations

- Provision scripts with mode `after-boot` or `ready` do not run on an automatic start (see open question 1). Warned about at enable time, and documented in the FAQ and the command help.
- The version check rejects development builds of Lima: `2.3.0-beta.0` and `2.3.0-17-gabc1234` both sort below `2.3.0` under semver. That is deliberate — it fails closed — but it does mean someone testing against Lima master is refused.

## LLM usage disclosure

Per `docs/CONTRIBUTE.md`. Drafted with Claude Code (Opus 5): `cmd/autostart.go`,
`environment/vm/lima/limautil/autostart.go`, and the FAQ addition were AI-generated from
my direction, then reviewed, corrected and tested by me before submitting. One defect
caught in review: the subcommands initially took a positional `[profile]` argument that
`cmd/root`'s whitelist would have silently ignored. I take responsibility for the code.


resker:
@abiosoft — before I invest any more into this, please advise:

This PR presently delegates to limactl autostart (merged but not yet shipped lima feature), which keeps the diff small but the generated unit runs limactl start rather than colima start — so after-boot/ready scripts don't run on an automatic start. Lima's plist hardcodes its own binary, so there's no way to point it at colima.

The alternative would be Colima generating its own LaunchAgent/LaunchDaemon (and systemd unit) running colima start -f. That'd give parity w/ what brew services does today at the cost of Colima owning plist generation, the sudo write, and launchctl bootstrap/bootout — though it'd also drop the Lima version dependency entirely.

I probably should have mentioned at the outset that the whole point is brew services can't cover the boot case w/o auto login. This PR would implement something that'd autostart on every boot regardless of whether logged in or not (e.g. on a headless system).

I suspect this initial version isn't sufficient given the gaps around after-boot/ready… but wanted to check prior to implementing the "parity" flavor of things.

PR #1628, opened 2026-08-26
net: add configurable subnet for shared networking

stek29:
Add network.subnet and --network-subnet for QEMU and Krunkit shared networking while preserving the existing 192.168.106.0/24 default. Derive and pass the socket_vmnet gateway, DHCP end, and mask from the configured IPv4 CIDR.

Refs #573

PR #1639, opened 2026-09-15
net: add configurable nat66Prefix for shared networking

stek29:
Add network.nat66Prefix and network-nat66-prefix for vmnet non-VZ shared networking, while preserving the existing behavior of not passing any value for nat66-prefix to vmnet by default.

This allows creating multiple colima instances with shared vmnet using IPv6 and is a follow-up for PR #1628, which fixed IPv4 use-case.

_LLM disclosure:_ I've used LLM based coding Agent for spinning up test VMs before/after this change to test different options. None of the code (including tests) was written by an LLM.


stek29:
example reproducer:
```
for n in 1 2; do
  colima start "shared-noprefix-v6-$n" \
    --vm-type qemu --network-address \
    --network-subnet "192.168.11$n.0/24" \
    # --network-nat66-prefix "fd22:cafe:babe:f00${n}::"
done

for n in 1 2; do
  colima ssh --profile "shared-noprefix-v6-$n" -- \
    curl -v --max-time 5 "https://[2606:4700:4700::1111]/"
done
```

here second VM's IPv6 will time out without the `network-nat66-prefix` fix.

abiosoft:
Is `nat66` a convention? I would expect the wording simply include `ipv6`.

stek29:
@abiosoft it's the name of the option in vmnet: https://developer.apple.com/documentation/vmnet/vmnet_nat66_prefix_key

abiosoft:
Kindly rebase.

stek29:
@abiosoft rebased, but actions seem to be failing because of a billing issue.

abiosoft:
> actions seem to be failing because of a billing issue.

Yeah, it was due to exceeding budget of copilot usage and an unusual rejection of a previously valid payment card.
It should get resolved within the next week.

abiosoft (approved):

majidamomajidam:
عالی هستی

